PR #595, opened 2023-07-25
Add Aria Labels to Confirmation Page

CalebPena:
- Add aria labels to confirmation page edit buttons.


ivonne-hernandez (approved):
LGTM 👍 .

PR #620, opened 2023-08-25
Convert externalId to string type, rm Num type

ivonne-hernandez:
- I converted the `externalId` to a string type to resolve the `null` field issue that's coming up on our dashboard.


CalebPena (approved):
LGTM

PR #621, opened 2023-08-30
Use Google Tag Manager

CalebPena:
- Use data layer instead of react-ga4.
- Send data layer event on page change.

### DIFF
--- a/src/Assets/analytics.ts
+++ b/src/Assets/analytics.ts
@@ -1,4 +1,4 @@
-export default function dataLayerPush(obj: any) {
+export default function dataLayerPush<T>(obj: T) {
   if (window.dataLayer === undefined) {
     console.error('Analytics are not working');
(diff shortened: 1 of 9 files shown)


src/Assets/analytics.ts
```diff
@@ -0,0 +1,7 @@
+export default function dataLayerPush(obj: any) {
```
ivonne-hernandez, line 1: What do you think about calling this `addToDataLayer`?

CalebPena: It is called a dataLayerPush in GTM.

ivonne-hernandez, line 1: What do you think about creating a function for each of the different param object types so that we don't have to declare the data type as `any` here?

ivonne-hernandez (approved):

PR #635, opened 2023-09-21
Refactor Referrer Code Logic

CalebPena:
- Skip referrer question if there is a referrer code.

- Refactor referrer code logic into a custom hook.


src/Assets/stepDirectory.ts
```diff
@@ -0,0 +1,51 @@
...
+  'acuteHHConditions',
+  'signUpInfo',
+];
+
+export function getStepDirectory(referrer: string | undefined) {
+  if (referrer) {
+    // if referrer is not undefinded or '' skip the referrer question
+    return referrerStepDirectory;
+  }
+  return defaultStepDirectory;
```
ivonne-hernandez, line 35: What do you think about wrapping this other return statement in an `else` block?

src/Components/Referrer/referrerHook.ts
```diff
@@ -0,0 +1,64 @@
...
+    default: 'https://www.myfriendben.org/',
+    '211co': 'https://screener.myfriendben.org?referrer=211co',
+  },
+};
+
+export default function useReferrer(referrerCode?: string) {
+  const [referrer, setReferrer] = useState<string | undefined>(referrerCode);
+
+  function getReferrer(key: keyof ReferrerData) {
+    return referrerData[key][referrer ?? 'default'] ?? referrerData[key]['default'];
```
ivonne-hernandez, line 60: If there is a `referrer` then it will look up the corresponding value in the dictionary but note that it may not have a value for that `referrer` in the dictionary which will make the entire `referrerData[key][referrer ?? 'default']` value `undefined` in which case it will use the second case `referrerData[key]['default']`.

ivonne-hernandez (approved):
This is fantastic :)!

PR #640, opened 2023-09-29
Define context via var destructuring

ivonne-hernandez:
- This will fix the following react-dom.production.min.js:186 ReferenceError: context is not defined error:
Screenshot 2023-09-29 at 9 44 33 AM


CalebPena (approved):
LGTM

PR #643, opened 2023-10-05
Rename Child Care Youth and Education Category

CalebPena:
- Renamed the preschool child care youth and education category.

### DIFF
--- a/src/Components/Results/Results.tsx
+++ b/src/Components/Results/Results.tsx
@@ -157,7 +157,7 @@ const Results = () => {
     });
   };
 
-  const preschoolProgramCategory = 'Cuidado de Niños, Preescolar y Jóvenes';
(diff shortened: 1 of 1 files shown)


src/Components/Results/Results.tsx
```diff
@@ -608,6 +608,14 @@ const Results = () => {
       );
     };
 
+    const currentCategory = categories.find((cat) => {
+      if (filt.category === false) {
+        return false;
+      }
+
+      return cat.defaultMessage === filt.category.value;
+    });
```
ivonne-hernandez, line 617: The [find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) array iterator method will either return the first element for which the statement is true or return `undefined`. What do you think about removing lines `612`-`614`?

ivonne-hernandez (approved):

PR #715, opened 2023-11-06
Navigator Label Translation

CalebPena:
- Add `Link: ` translation.
- Add `Email: ` translation.
- Add `Phone Number: ` translation.

### DIFF
--- a/src/Components/Results/Results.tsx
+++ b/src/Components/Results/Results.tsx
@@ -335,7 +335,7 @@ const Results = ({ handleTextFieldChange }: ResultsProps) => {
                   </p>
                   {navigator.assistance_link.default_message && (
                     <h4 className="font-weight">
-                      Link:{' '}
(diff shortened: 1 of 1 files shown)


src/Components/Results/Results.tsx
```diff
@@ -335,7 +335,7 @@ const Results = ({ handleTextFieldChange }: ResultsProps) => {
                   </p>
                   {navigator.assistance_link.default_message && (
                     <h4 className="font-weight">
-                      Link:{' '}
+                      <FormattedMessage defaultMessage="Link: " id="navigator.linkLabel" />
```
ivonne-hernandez, line 338: What do you think about having the `id` listed first and then the `defaultMessage` on all of these?

ivonne-hernandez (changes requested):
What do you think about having the `id` listed first and then the `defaultMessage` on all of these to keep it consistent with the rest of the codebase?

civicpixel (approved):

PR #717, opened 2023-11-07
Feat/page title tags

ivonne-hernandez:
ClickUp Ticket: https://app.clickup.com/t/8685zwv3r

- I dynamically set the `document.title` for each page based on the `location`.

- I converted the `startingQuestionNumber` variable into a magic number so that it's easier to distinguish them from our regular variables.

- The biggest concern that I have in regards to this feature is that if the step order changes in the `stepDirectory.ts` file, then we may need to refactor the way that the `document.title` is being set for each page since it will not match up with the `pageTitleTags`.

Screenshot 2023-11-07 at 4 31 10 PM
Screenshot 2023-11-07 at 4 31 17 PM
Screenshot 2023-11-07 at 4 31 26 PM
Screenshot 2023-11-07 at 4 31 35 PM
Screenshot 2023-11-07 at 4 31 49 PM
Screenshot 2023-11-07 at 4 32 13 PM
Screenshot 2023-11-07 at 4 32 20 PM
etc...


src/Assets/pageTitleTags.ts
```diff
@@ -0,0 +1,17 @@
+const pageTitleTags: Record<string, string> = {
+  'step-1': 'Preferred Language',
```
CalebPena, line 2: Do we want to do `My Friend Ben | Preferred Language`.

src/Assets/pageTitleTags.ts
```diff
@@ -0,0 +1,17 @@
+const pageTitleTags: Record<string, string> = {
```
CalebPena, line 1: With Records, you want to include all of the keys as a union type. You could also just let the type be inferred because this is a constant.
Ex:
```
type StepName = 'step-1' | 'step-2' | ...
const pageTitleTags: Record<StepName, string> = {
```

CalebPena, line 1: If you want to follow the constant naming convention, `pageTitleTags` is also a constant.

CalebPena (approved):

src/Assets/pageTitleTags.ts
ivonne-hernandez: I think that just having the `Preferred Language` is fine because otherwise the `My Friend Ben | Preferred Language` string just runs off the tab:
Screenshot 2023-11-09 at 11 44 57 AM

PR #729, opened 2023-11-09
Refactor/childcare cat header value

ivonne-hernandez:
- I refactored the `renderAllCategoryValues` and `useEffect` functions in `Results.tsx` by capping the preschool and child care total. 

- No issues arose but what do we think about breaking out the Youth and Education Programs like the Pell Grant, My Spark and My Denver Card into their own category? This would help make capping much more straight forward instead of having this `const isPreschoolOrChildCareProgram = ['upk', 'dpp', 'chs', 'cccap'].includes(program.short_name)` hard coded condition that will need to be updated every time we add a new childcare or preschool program.

New `Child Care, Youth, and Education` category header on desktop:
Screenshot 2023-11-09 at 10 47 50 AM

New `Child Care, Youth, and Education` category header on mobile:
Screenshot 2023-11-09 at 11 00 53 AM

### DIFF
--- a/src/Components/Results/Results.tsx
+++ b/src/Components/Results/Results.tsx
@@ -251,7 +251,8 @@ const Results = ({ handleTextFieldChange }: ResultsProps) => {
   const preschoolProgramCategoryString = 'Child Care, Youth, and Education';
 
   const renderAllCategoryValues = (programs: Program[]) => {
-    const childCareYouthEdPrograms = { childCareTotalEstVal: 0, youthAndEducationTotalEstVal: 0 };
+    let childCareTotalEstVal = 0;
+    let youthAndEducationTotalEstVal = 0;
     const categoryValues: { [key: string]: number } = {};
 
     for (let program of programs) {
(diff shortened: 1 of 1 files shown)


src/Components/Results/Results.tsx
```diff
@@ -265,18 +272,24 @@ const Results = ({ handleTextFieldChange }: ResultsProps) => {
...
+          isPreschoolOrChildCareProgram === false
+        ) {
+          youthAndEducationTotalEstVal += program.estimated_value;
         }
       }
     }
 
-    if (preschoolPrograms.totalEstVal > 8640 && preschoolPrograms.numOfPreSchoolPrograms > 1) {
-      categoryValues[preschoolProgramCategoryString] = 8640;
+    if (childCareTotalEstVal > 8640) {
```
CalebPena, line 287: Did you mean to remove the logic for if there is one preschool program to not cap the value?

CalebPena (approved):

PR #763, opened 2023-11-28
Set app language based on browser default

myang1010:
- allow the application to see what language by browser is set to and default to that language if it is available
- added `userLanguage` to Header.js and SelectLanguagePage.tsx. If the user's browser language matches our language options, set the app language accordingly; otherwise, default to English.

### DIFF
--- a/src/Components/Header/Header.js
+++ b/src/Components/Header/Header.js
@@ -25,6 +25,7 @@ const Header = ({ handleTextfieldChange }) => {
   const [openShare, setOpenShare] = useState(false);
   const [openEmailResults, setOpenEmailResults] = useState(false);
   const [isLanguageSelectOpen, setIsLanguageSelectOpen] = useState(false);
+  const userLanguage = navigator.language?.toLocaleLowerCase();
 
   const handleOpenShare = () => {
     setOpenShare(true);
@@ -51,7 +52,7 @@ const Header = ({ handleTextfieldChange }) => {
   };
 
   const setRenderValue = () => {
-    const currentLocale = context.locale;
+    const currentLocale = (userLanguage in languageOptions) ? userLanguage : context.locale
     return currentLocale.slice(0, 2).toLocaleUpperCase();
   };
 
@@ -80,7 +81,7 @@ const Header = ({ handleTextfieldChange }) => {
             <Select
               labelId="select-language-label"
               id="select-language"
-              value={context.locale}
+              value={userLanguage}
               label="Language"
               onChange={(event) => context.selectLanguage(event)}
               aria-label="select a language"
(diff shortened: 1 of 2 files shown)


src/Components/Header/Header.js
```diff
@@ -80,7 +81,7 @@ const Header = ({ handleTextfieldChange }) => {
             <Select
               labelId="select-language-label"
               id="select-language"
-              value={context.locale}
+              value={userLanguage in languageOptions ? userLanguage : context.locale}
```
CalebPena, line 84: This does not change the language, it just changes what the select is showing. To set the language, update the `Wrapper.tsx` file, and have the default language on line 103 be set to local storage if it exists, or to the navigator language.

src/Components/Header/Header.js
```diff
@@ -25,6 +25,7 @@ const Header = ({ handleTextfieldChange }) => {
   const [openShare, setOpenShare] = useState(false);
   const [openEmailResults, setOpenEmailResults] = useState(false);
   const [isLanguageSelectOpen, setIsLanguageSelectOpen] = useState(false);
+  const userLanguage = navigator.language?.toLocaleLowerCase();
```
CalebPena, line 28: Some users might have a region code attached to their language code. For example, if a user has a language code of `es-ar` that means that they Speak Spanish from Argentina, but we would still want to show them MFB in Spanish. Also, `.toLocaleLowerCase()` can be different depending on the users language, so it is better to use `.toLowerCase()` here in order to prevent any weird bugs. More info [here](https://stackoverflow.com/questions/34283414/difference-between-tolocalelowercase-and-tolowercase)

CalebPena (changes requested):

src/Components/Wrapper/Wrapper.tsx
```diff
@@ -100,19 +100,23 @@ const Wrapper = (props: PropsWithChildren<{}>) => {
       setTranslations(value);
     });
   }, []);
-  let defaultLanguage = (localStorage.getItem('language') ?? 'en-us') as Language;
+  const defaultLanguage = (localStorage.getItem('language') ?? 'en-us') as Language;
+
+  let userLanguage = navigator.language.toLowerCase() as Language;
+  userLanguage = languageOptions[userLanguage] ? userLanguage : defaultLanguage;
```
CalebPena, line 106: We want to set the language to the `localStorage` language if it exists, and if it does not exist, we want to set the language to the browser default. This is because we do not want to change the language if the user refreshes the page. Also to check if the navigator language exists, we want to do something like what is shown bellow to account for region codes:

```
Object.keys(languageOptions).some((lang) => userLanguage.slice(2) === lang.slice(2))
```

CalebPena (changes requested):

src/Components/Wrapper/Wrapper.tsx
myang1010: Thanks for clarification. Now, we check the `localStorage` for the preferred language. If it's not specified, we then check the user's browser language. If none of them is available, we set the default language to `en-us`.

CalebPena (approved):
Looks good.

PR #808, opened 2023-12-12
Email Error

CalebPena:
- Add an error message if the server throws an error for email address.

- We can not currently determine the exact error with the email, so we have to use a generic error message. @sydney-devine Let me know if you would like to change any of the wording on the error message.

image


sydney-devine:
@CalebPena 

Some typos and updated language. 

> Please enter a valid email address. This error could also be caused by entering an email address that is already in our system. If the error persists, remember that this question is optional and will not impact your MyFriendBen results. You can skip this question by deselecting the boxes at the top of the page and pressing continue. 

Can we align the error icon with the first line?

civicpixel (approved):

PR #813, opened 2023-12-14
Rebrand Adjust Back and Continue buttons

myang1010:
- Added 1rem margin top to `.question-buttons`
- Removed shadow for Continue buttons

- Do we want to maintain the same style for the `Add Another Income` button as the "+Add Another Expense" button in step 6?


CalebPena (approved):
Looks good

Issue #825, opened 2023-12-19
Create updated header for results page


sydney-devine:
Background Color: #ECDEED 
Background for # Programs Found: #293457
Text Color: #293457
Font: Roboto Slab for $ values | Open Sans for label underneath

https://xd.adobe.com/view/9c25b118-90b5-409e-b9b6-36662f2b59c2-c976/?fullscreen

![image](https://github.com/Gary-Community-Ventures/benefits-calculator/assets/109622199/052015ac-d397-46ae-b64d-6c1ba64f0958)

rickyctd:
There are some errors on this page in the backend that prevent this from displaying in local.

sydney-devine:
@CalebPena ^ see Ricky's note about results page in local. Can we get this figured out for their team?

rickyctd:
@sydney-devine @CalebPena I think @mexi-cano was able to figure it out on his local and working with the rest of our team to figure it out on their locals. They will probably reach out if additional help is needed. Might want consider seeing if this is a more systemic issue that should be documented/resolved for future devs though?

Mdemenko1:
@sydney-devine maybe Background Color, should be #FBF9FC (secondaryBackgroundColor) ?

sydney-devine:
> 

Agree. I think I grabbed the wrong hex code when I typed this. It should definitely be #FBF9FC - sorry about that.

PR #862, opened 2024-01-17
825/header result page 

Mdemenko1:

Work for the issue [825](https://github.com/Gary-Community-Ventures/benefits-calculator/issues/825)

Creating an updated header for the result page

Look:

image

i-icon has a hover state as per design https://xd.adobe.com/view/9c25b118-90b5-409e-b9b6-36662f2b59c2-c976/?fullscreen

Created function `calculateTotalValue `to calculate value based on a category passed


Mdemenko1:
@sydney-devine let me know if size of the header/text and etc looks ok to you

src/Components/Results/Header/Header.tsx
```diff
@@ -15,8 +18,34 @@ const Buttons = () => {
...
+  return (
+    <CardContent sx={{ backgroundColor: theme.secondaryBackgroundColor }}>
+      <div className="results-header">
+        <div className="results-header-programs-count-text">
+          <div className="results-header-programs-count">{programs.length}</div>
+          <div>Programs Found</div>
+        </div>
+        <div style={{ marginLeft: '15px' }}>
+          <div className="results-header-values">
+            ${Math.round(estimatedMonthlySavings / 12).toLocaleString()} <div className="info-circle">i</div>
```
CalebPena, line 38: We can remove the help icons for now because that feature is not implemented yet.
```suggestion
            ${Math.round(estimatedMonthlySavings / 12).toLocaleString()}
```

CalebPena (approved):
Looks good to me, my only question is for @sydney-devine if we want the header to look different for desktop?

sydney-devine:
> Looks good to me, my only question is for @sydney-devine if we want the header to look different for desktop?

@Mdemenko1 team chatted about this one this morning since it does just have way too much blank space on desktop. Can we make the desktop and tablet version varied to look like this mock @kc-accidental did? 

image

sydney-devine (changes requested):
See comment re desktop/tablet version.

src/Components/Results/Header/Header.tsx
```diff
@@ -15,8 +18,30 @@ const Buttons = () => {
...
+        <div className="results-header-programs-count-text">
+          <div className="results-header-programs-count">{programs.length}</div>
+          <div>Programs Found</div>
+        </div>
+        <div style={{ marginLeft: '15px' }}>
+          <div className="results-header-values">${Math.round(estimatedMonthlySavings / 12).toLocaleString()}</div>
+          <div className="results-header-label">Estimated Monthly Savings</div>
+          <div className="results-header-values">${taxCredit}</div>
+          <div className="results-header-label">Estimated Tax Credit</div>
+        </div>
```
ivonne-hernandez, line 41: What do you think about using more semantic html tags for some of these elements? We want to make sure that we're making our website as accessible as we possibly can by using semantic html tags.

Mdemenko1: @ivonne-hernandez  Please check the updated code, let me know if you have something specific what would you like to substitute with

CalebPena (approved):
Looks good.

sydney-devine (approved):
Updated visual looks great on my end.

PR #876, opened 2024-01-23
Add Documents to Program Type

CalebPena:
- Add documents to program type.

### DIFF
--- a/src/Types/Results.ts
+++ b/src/Types/Results.ts
@@ -36,6 +36,7 @@ export type Program = {
   already_has: boolean;
   new: boolean;
   navigators: ProgramNavigator[];
+  documents: Translation[];
 };
 
 export type UrgentNeed = {


src/Types/Results.ts
```diff
@@ -36,6 +36,7 @@ export type Program = {
   already_has: boolean;
   new: boolean;
   navigators: ProgramNavigator[];
+  documents: Translation[];
```
myang1010, line 39: Thanks for the change. I know where to find documents now. Just one question, I guess I have to wait for the backend change get merged first then I can see this new field, right? Thanks.

CalebPena: Yes, that change will be merged this Friday.

ivonne-hernandez (approved):

ivonne-hernandez (approved):

PR #879, opened 2024-01-24
Create program page

myang1010:
- Setup the program detail page

- Removed a comma in Review page

- Not sure where does the Get Help, Apply Now should redirect to
- Also the `Save this result` button doesn't have any function in it

- `Estimated Annual Value` and `Estimated Time to Apply` text, and  `Get Help`, `Apply Now` and `SAVE THIS RESULT` button names currently don't have any translation.

Screenshots

- desktop, large screen
Screenshot 2024-01-31 at 2 03 08 PM

- mobile, small screen
Screenshot 2024-01-31 at 1 55 39 PM


sydney-devine (changes requested):
I'm the worst. Realizing that the link in the ticket was the original wireframe link and we had updated it slightly. This is totally my bad, and I'm so sorry I didn't catch that. 

https://xd.adobe.com/view/9c25b118-90b5-409e-b9b6-36662f2b59c2-c976/screen/2341183d-f7d8-4568-8ec9-32c96d71b5c3/?fullscreen

Changes visually from what I'm seeing:
- Where it says Medicaid in the orange at the top, that should be pulling from the category heading. Want to make sure it's doing that. 
- Remove the "Get Help" button under the estimated value/time to apply box
- Add in Get Help Applying box, light lavender background. The information that pulls into this box should come from the navigators info on the backend. @CalebPena can you help specify where that might be found? 
- Save my results button should go back on top right. But I think that might be addressed with the work that @ivonne-hernandez is doing for those top navigational buttons. 
- The current "Back to Results" button, can we left alight it with the estimated annual value / time to apply box so it isn't floating so far to the left?

CalebPena:
Navigators is the `program.navigators` field.

myang1010:
@sydney-devine and @CalebPena thanks for the feedback. A few more questions regarding the updated design doc, 

Screenshot 2024-01-24 at 3 39 08 PM

1. The estimated time we have by default is like `45 minutes` while the design doc is `45 Min`, do we want exact match with the design doc?
2. Do we have the question mark icon that I can use?
3. For the `program.navigators` seems they'll come in an array. Do we know the data format of each element inside the navigators array? Looks like there there will be some sort of organization name, address and contact info. Is it possible that I can get some mock data for this field(Not a blocker, but nice to have)?

Thanks.

sydney-devine:
> @sydney-devine and @CalebPena thanks for the feedback. A few more questions regarding the updated design doc,
> 
> Screenshot 2024-01-24 at 3 39 08 PM
> 1. The estimated time we have by default is like `45 minutes` while the design doc is `45 Min`, do we want exact match with the design doc?
> 2. Do we have the question mark icon that I can use?
> 3. For the `program.navigators` seems they'll come in an array. Do we know the data format of each element inside the navigators array? Looks like there there will be some sort of organization name, address and contact info. Is it possible that I can get some mock data for this field(Not a blocker, but nice to have)?
> 
> Thanks.

Hi @myang1010 

1. I would say let's go with "min". I believe that's something we can control on the backend by updating those values to just say 45 min though. @CalebPena can you confirm that's something we can just update? 
2. Ignore the question mark icon for now. It's going to go with the help bubble ticket. That's just mocked up there as a potential for what it could look like. 
3. @CalebPena are you able to provide this for Ming?

CalebPena:
@myang1010 The estimated time is a string sent from the backend, so we can just update it in the backend. Also, you can get some mock navigator data if you go to Django admin, go to the navigators tab and click on a navigator, and then select some programs for that navigator.

image

CalebPena:
FYI, the Results top buttons pr is merged, so you can use it in this pr.

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,104 @@
...
-    </>
+
+      <section className="apply-box">
+        <h3>Get Help Applying</h3>
+        <ul className="apply-box-list">
+          {program.navigators.map((info, index) => (
+            <li key={index} className="apply-info">
+              <ResultsTranslate translation={info.name} />
+              <address className="address-info">
+                <p>Address bla bla bla</p>
```
ivonne-hernandez, line 75: What do you think about adding the `info.address` here instead of `Address bla bla bla`?

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,104 @@
...
+                <a href={info.assistance_link.default_message}>Visit Website</a>
```
ivonne-hernandez, line 76: Can you please wrap the `Visit Website` text in a `FormattedMessage`?

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,104 @@
...
+                <br />
+                <a href={`mailto:${info.email}`}>{info.email.default_message}someone@somewhere.com</a>
+                <br />
+                <a href={`tel:${info.phone_number}`}>{info.phone_number}123.123.1233</a>
```
ivonne-hernandez, line 80: Same here. What do you think about removing `123.123.1233`?

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,104 @@
...
+              </address>
+            </li>
+          ))}
+        </ul>
+      </section>
+
+      <section className="required-docs">
+        <h3>Required Documents Checklist</h3>
```
ivonne-hernandez, line 88: Can you please wrap this text in a `FormattedMessage`?

ivonne-hernandez (changes requested):
Please see comments above for requested changes.

myang1010:
Thank you @ivonne-hernandez and @CalebPena for the feedback.

1. I've removed those unnecessary text for email and phone number. 

   I'm keeping the address info as I don't see the `address` field from `navigator`. Please let me know if I'm using the wrong input. Here's what I have, 

    >Screenshot 2024-02-08 at 2 18 40 PM

2. I've wrapped "Visit Website" text and "Required Documents Checklist" text in FormattedMessage. But I'm not sure about the correct ids for those two. I'm giving them `id="results.visit-webiste"` and `id="results.required-documents-checklist"` for now. Also unsure about id for "BACK TO RESULTS" button, so giving it `id="results.back-to-results-btn"`.
3. I also noticed the "Warning(69:3) autoprefixer: start value has mixed support, consider using flex-start instead". It's from `.MuiFormControlLabel-root.vertical-align` class in Filter.css. If it hasn't been taken cared of, I can change it to `flex-start` in this PR.

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
+          {program.navigators.map((info, index) => (
+            <li key={index} className="apply-info">
+              <ResultsTranslate translation={info.name} />
+              <address className="address-info">
+                <p>Address info goes here</p> {/* replace with info.address */}
+                <a href={info.assistance_link.default_message}>
+                  <FormattedMessage id="results.visit-webiste" defaultMessage="Visit Website" />
+                </a>
+                <br />
+                <a href={`mailto:${info.email}`}>{info.email.default_message}</a>
```
CalebPena, line 87: We support having a different email depending on the language, could you translate `info.email`

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
       </div>
-    </>
+
+      <section className="apply-box">
+        <h3>Get Help Applying</h3>
+        <ul className="apply-box-list">
+          {program.navigators.map((info, index) => (
+            <li key={index} className="apply-info">
+              <ResultsTranslate translation={info.name} />
+              <address className="address-info">
```
CalebPena, line 81: Why did you put all of the information in the `address` element? I am pretty sure that the `address` element is should only be used for actual addresses

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
+                <p>Address info goes here</p> {/* replace with info.address */}
+                <a href={info.assistance_link.default_message}>
```
CalebPena, line 83: `email`, `phone`, and `assistance_link` can be blank, if a navigator does not have one of those things. Could you make it so that if there is no default message, then the field does not show up.

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
+        <h3>
+          <FormattedMessage id="results.required-documents-checklist" defaultMessage="Required Documents Checklist" />
+        </h3>
+        <ul className="required-docs-list">
+          {program.documents.map((document, index) => (
+            <li key={index}>
+              <ResultsTranslate translation={document} />
+            </li>
+          ))}
+        </ul>
```
CalebPena, line 106: Documents can be an empty array, could you not render this if there are no documents.

CalebPena (changes requested):

sydney-devine (changes requested):
Can we update the desktop version of it so that line between the category name and the benefit name doesn't extend so far past everything else?

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
+          <div className="estimation-text-left">Estimated Time to Apply</div>
+          <div className="estimation-text-right">
+            <ResultsTranslate translation={program.estimated_application_time} />
+          </div>
+        </div>
+      </section>
+
+      <div className="apply-online-button">
+        <Link to={program.apply_button_link.default_message} target="_blank" rel="noopener noreferrer">
+          Apply Online
```
CalebPena, line 71: All plane text should be in a `FormattedMessage`, so that it can be translated.

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -1,24 +1,115 @@
...
-    </>
+
+      <section className="apply-box">
+        <h3>Get Help Applying</h3>
+        <ul className="apply-box-list">
+          {program.navigators.map((info, index) => (
+            <li key={index} className="apply-info">
+              <ResultsTranslate translation={info.name} />
+              <address className="address-info">
+                <p>Address info goes here</p> {/* replace with info.address */}
```
CalebPena, line 82: I think that we can get rid of this for now, because we don't currently have a backend field for addresses.

CalebPena (changes requested):

myang1010:
@CalebPena, in this latest commit, I've converted all measurements from pixels (px) to rems (rem). I've also made adjustments so that navigators or documents are only rendered when there's a 'default_message' available. Additionally, I've wrapped all plain text within either FormattedMessage or ResultsTranslate components. Also website URLs, email addresses, and phone numbers can be placed within an `address` tag, although it's not strictly necessary. Replaced it with regular `div`. 

@sydney-devine , here's the updated screenshots,

no_nav_doc
only_doc
small_screen

Thank you all for the feedback. Please let me know if there's any other concerns. Thanks!

sydney-devine (approved):
Visuals look good on my end

Issue #857, opened 2024-01-16
Django regional frontend


sydney-devine:
## Create the Config Hook
Create a hook to add to the `Wrapper.tsx` component to fetch and return the typed config data.

```
// define the config type
type Config = {
  zipcode: any; // zipcode type,
  otherFields: any; // those types as well
};

async function getConfig() {
  // fetch data
  const data = {};
  return data as Config;
}

function useGetConfig() {
  const [loading, setLoading] = useState<boolean>(true);
  const [config, setConfig] = useState<Config | undefined>();

  useEffect(() => {
    getConfig().then((value) => {
      // get data and set loading to false
      setConfig(value);
      setLoading(false);
    });
  }, []);

  return { loading, config };
}
```

## Update the Wrapper and Context
Add the new hook to the wrapper
Update the page is loading logic to also wait for the config to load
Update the context to include the config

You should be able to use the following in other components.
```
const { config } = useContext(Context)
console.log(config.zipcode)
```

sydney-devine:
@CalebPena to add details in

PR #883, opened 2024-01-25
Created a Config Hook

mexi-cano:
- Created a hook to fetch and return the typed config data and added it to the Wrapper.tsx component

issue #857 

Screenshot 2024-01-25 at 10 54 04 AM


src/Types/ApiFormData.ts
CalebPena, line 1: These are not quite the types that we are going to use, but it is a good example. Can we move these into their own `Config.ts` file.

src/Components/LandingPage/LandingPage.tsx
```diff
@@ -29,7 +29,7 @@ interface LandingPageProps {
 }
 
 const LandingPage = ({ handleCheckboxChange }: LandingPageProps) => {
-  const { formData, locale, screenDoneLoading, theme } = useContext(Context);
+  const { formData, locale, screenDoneLoading, theme, config } = useContext(Context);
```
CalebPena, line 32: Because this is not being used yet, can we remove it from here for now.

src/Components/Wrapper/Wrapper.tsx
```diff
@@ -79,19 +83,27 @@ const initialFormData: FormData = {
 export const Context = React.createContext<WrapperContext>({} as WrapperContext);
 
 const Wrapper = (props: PropsWithChildren<{}>) => {
+  const configResponse = useGetConfig();
```
CalebPena, line 86:
```suggestion
  const { configLoading, configResponse } = useGetConfig();
```

src/Components/Wrapper/Wrapper.tsx
```diff
@@ -79,19 +83,27 @@ const initialFormData: FormData = {
...
   const [pageIsLoading, setPageIsLoading] = useState<boolean>(true);
+  const [configIsLoading, setConfigIsLoading] = useState<boolean>(true);
 
   const screenDoneLoading = () => {
     setScreenLoading(false);
   };
 
   useEffect(() => {
-    if (!screenLoading && !translationsLoading) {
+    if (!screenLoading && !translationsLoading && !configIsLoading) {
```
CalebPena, line 99:
```suggestion
    if (!screenLoading && !translationsLoading && !configLoading) {
```

src/Components/Config/configHook.tsx
```diff
@@ -0,0 +1,32 @@
...
+  });
+}
+
+export default function useGetConfig() {
+  const [loading, setLoading] = useState<boolean>(true);
+  const [config, setConfig] = useState<ApiConfig | undefined>();
+
+  useEffect(() => {
+    getConfig().then((value) => {
+      // get data and set loading to false
```
CalebPena, line 25: Can you loop through the api response, and create an object where the key is the name, and the value is the data.

CalebPena (changes requested):

CalebPena:
This is not a results redesign ticket, so can you rebase it on the dev branch.

src/Components/Config/configHook.tsx
mexi-cano: Would accessing the data as the example below not give us the needed object?

`const configData = Array.isArray(value) && value[0]?.data;`

src/Types/ApiFormData.ts
mexi-cano: This comment shows as `Outdated`, just making sure you still recommend moving the type to `Config.t`s?

CalebPena: Yes, please move the type to `Config.ts`. FYI, comments are marked as `Outdated` if the file they are in changes.

src/Components/Config/configHook.tsx
CalebPena: This requires all of the config to be in one model on the backend. The config response should look like:

```
[
  {
    "name": "zipcode",
    "data": [80205, zipcodes...]
  },
  {
    "name": "other",
    "data": {"other": "data"}
  }
]
```
and then be converted to:
```
{
  "zipcode": [80205, zipcodes...],
  "other": {"other": "data"}
}
```

PR #897, opened 2024-01-29
Bug/Remove colon for income and expense questions

myang1010:
- Removed the colon before the question mark for income and expense questions

- Someone with the translation key has to help remove the colon from each languages as well. 

Screenshots
- Income you
>income_you

- Incom they
>income_they

- Expense
>expense

### DIFF
--- a/src/Components/ExpenseBlock/ExpenseQuestion.js
+++ b/src/Components/ExpenseBlock/ExpenseQuestion.js
@@ -96,15 +96,11 @@ const ExpenseQuestion = ({ expenseData, allExpensesData, setAllExpenses, deleteE
   };
 
   const getExpenseSourceLabel = (expenseSourceName) => {
-    return expenseSourceName ? (
-      <>
-        {'('}
-        {expenseOptions[expenseSourceName]}
-        {')?'}
-      </>
-    ) : (
-      '?'
-    );
+    if (expenseSourceName) {
+      return `(${expenseOptions[expenseSourceName]})?`;
+    }
+
+    return '?';
   };
 
   const textfieldProps = {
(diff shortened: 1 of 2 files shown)


src/Components/ExpenseBlock/ExpenseQuestion.js
```diff
@@ -96,7 +96,15 @@ const ExpenseQuestion = ({ expenseData, allExpensesData, setAllExpenses, deleteE
...
-    return expenseOptions[expenseSourceName];
+    return expenseSourceName ? (
+      <>
+        {'('}
+        {expenseOptions[expenseSourceName]}
+        {')?'}
+      </>
+    ) : (
+      '?'
+    );
```
CalebPena, line 107: Guard clauses are usually easier to read than ternaries. Also I refactored the expense option part to just return a string instead of JSX.
```suggestion
    if (expenseSourceName) {
      return `(${expenseOptions[expenseSourceName]})?`;
    }
    
    return '?';
```

src/Components/IncomeBlock/IncomeQuestion.js
```diff
@@ -59,7 +59,15 @@ const IncomeQuestion = ({
...
-    return incomeOptions[incomeStreamName];
+    return incomeStreamName ? (
+      <>
+        {'('}
+        {incomeOptions[incomeStreamName]}
+        {')?'}
+      </>
+    ) : (
+      '?'
+    );
```
CalebPena, line 70: Can you use a guard clause here?

CalebPena (approved):
Looks good, just a couple of suggestions to make it more readable.

src/Components/ExpenseBlock/ExpenseQuestion.js
myang1010: Thanks for the suggestion. Currently if I just return as string directly, like this ``return `(${expenseOptions[expenseSourceName]})?`;``, I got [object Object]. Let me know if I need to temporarily return as JSX.

CalebPena: I did not realize that `expenseOptions[expenseSourceName]` returned a `FormattedMessage`. Here is an updated suggestion.
```suggestion
    if (expenseSourceName) {
      return <>({expenseOptions[expenseSourceName]})?</>;
    }
    
    return '?';
```

CalebPena (approved):
Looks good.

PR #914, opened 2024-02-02
Warning Icon

CalebPena:
- Move `svg` icons to their own file.
- Change `BackToScreen` warning icon.


src/Components/BackToScreen/BackToScreen.css
```diff
@@ -12,3 +12,10 @@
   font-size: 1rem;
   cursor: pointer;
 }
+
+.back-to-screen-warning-icon {
+  min-width: 2rem;
+  min-height: 2rem;
```
ivonne-hernandez, line 18: What do you think about removing these properties since the width and height are hard coded to be 2rems below?

ivonne-hernandez (approved):

src/Components/BackToScreen/BackToScreen.css
CalebPena: I tried that, but it did not work.

PR #920, opened 2024-02-06
Created component for near-term resources cards

mexi-cano:
- Created component for near-term resources cards
- Existing acute conditions object descriptions differed from backend. Created new object for mapping.

Screenshot 2024-02-06 at 1 39 56 PM

Screenshot 2024-02-06 at 1 40 16 PM


src/Components/Results/Needs/NeedCard.tsx
```diff
@@ -1,14 +1,52 @@
...
+      <div className="result-resource-more-info">
+        {urgentNeed.icon}
+        <span>
+          <ResultsTranslate translation={translateNeed} />
+          <strong>
+            <ResultsTranslate translation={need.name} />
+          </strong>
+        </span>
+        <a className="resource-more-info" href={needLink} target="_blank" rel="noreferrer">
+          More Info
```
CalebPena, line 47: Make this a `FormattedMessage`

src/Components/Results/Needs/NeedCard.tsx
```diff
@@ -1,14 +1,52 @@
...
+    return { icon, defaultMessage: default_message };
+  }
+
+  return { icon: null, defaultMessage: '' };
+};
+
 const NeedCard = ({ need }: NeedsCardProps) => {
+  const needType = need.type.default_message;
+  const urgentNeed = getIconAndDefaultMessage(needType, acuteConditionResultMapping);
+  const translateNeed = { default_message: urgentNeed.defaultMessage, label: urgentNeed.defaultMessage };
```
CalebPena, line 34: Isn't this just `need.type`

src/Assets/acuteConditionOptions.tsx
```diff
@@ -9,6 +9,55 @@ import { ReactComponent as Job_resources } from './OptionCardIcons/AcuteConditio
 import { ReactComponent as Legal_services } from './OptionCardIcons/AcuteConditions/legal_services.svg';
 import { ReactComponent as Support } from './OptionCardIcons/AcuteConditions/support.svg';
 
+export const acuteConditionResultMapping = {
```
CalebPena, line 12: Because the urgent needs have the type text already, I think that we can just make this an icon map. Also, I think we should use a different class name because this is not being used in an option card.

Example: 
```
{
  'Food or groceries': <Food className="urgent-need-icon" />,
  ...
}
```

src/Components/Results/Needs/NeedCard.tsx
```diff
@@ -1,14 +1,52 @@
...
 };
 
+interface ResponseObject {
+  icon: React.ReactNode | null;
+  defaultMessage: string;
+}
+
+const getIconAndDefaultMessage = (
+  messageType: string,
+  acuteConditionResultMapping: Record<string, any>,
```
CalebPena, line 17: We don't need to pass in this as a prop, since we are importing it.

CalebPena (changes requested):

src/Assets/acuteConditionOptions.tsx
mexi-cano: The issue is that the type text is too long to use as headers. I created `acuteConditionResultMapping` to hold the heading text and `const translateNeed` to pass said text to `ResultTranslate`

sydney-devine (approved):
Visual looks great - TY!

src/Assets/acuteConditionOptions.tsx
CalebPena: I think that we should use the type text from the API, and then we can shorten the type text on the urgent need type.

@kbottagaro Can we update the urgent need types when we release this change.

kbottagaro:
@Caleb Peña ***@***.***> I did share this doc
<https://docs.google.com/document/d/1QXpbMxjW3u4f7qhiSAMEwNCMlphZzPj40c_tQFkGyDA/edit?usp=sharing>
the other day with Sydney and Carlos. Do I need to update these in the
database today? And if so, before or after releases?

src/Assets/acuteConditionOptions.tsx
mexi-cano: Ok, I'll change it to use the type text from the API. In order to map it to the corresponding icon, I believe `acuteConditionResultMapping` will still need to hold the icon and the new shortened type text.

CalebPena: The type text can be changed in the database.

If you go to http://localhost:8000/api/translations/admin/urgent_needs
You can then select whichever program you want, and then select the type link, and edit the type text.

mexi-cano: > The type text can be changed in the database.
> 
> If you go to http://localhost:8000/api/translations/admin/urgent_needs You can then select whichever program you want, and then select the type link, and edit the type text.

I updated the backend type texts locally. See screenshots below. 

Screenshot 2024-02-13 at 2 48 09 PM

Screenshot 2024-02-13 at 2 43 23 PM

CalebPena (approved):
LGTM

Issue #869, opened 2024-01-18
Update Prettier to Catch Console Logs


CalebPena:
Update Prettier so that it won't let prs be merged if there is a `console.log`

rickyctd:
@Mdemenko1 will add instructs on PR

Consider creating new ticket to improve vis github actions.

PR #921, opened 2024-02-06
869: Validate files for console.log statement

Mdemenko1:
 Work for the issue [869](https://github.com/Gary-Community-Ventures/benefits-calculator/issues/869)
- Feature to validate staged files for console.log() statements
- Added dependencies:
  -   husky: "^9.0.10",
  -   lint-staged: "^15.2.2"

 - Changes above will replace need to run `npm run format` since this script will be executed at a time of committing changes
 - When console.log("lorem ipsum") is present in the staged files eslint will return error: 
 `error  Unexpected console statement  no-console`
 
image

## Any other comments, questions, or concerns?
- run: `npm i` to install latest updates
- need to specify `eslint-plugin-react` version in the `package.json`


package.json
```diff
@@ -74,6 +74,19 @@
...
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "src/**/*.{js,jsx,ts,tsx}": [
+      "prettier --write",
+      "eslint --fix"
+    ]
```
ivonne-hernandez, line 90: I appreciate your work on this. Could we look into configuring the script to run during pull requests instead of commits? This way, we can avoid adding extra dependencies and save time during development by ensuring it runs only when necessary. Perhaps we could start by keeping things simple for now and focus on maintaining the script within our existing setup by keeping `Line 16` and `Line 17` in .`eslintrc.json`.

ivonne-hernandez (changes requested):
I appreciate your work on this. Could we look into configuring the script to run during pull requests instead of commits? This way, we can avoid adding extra dependencies and save time during development by ensuring it runs only when necessary. Perhaps we could start by keeping things simple for now and focus on maintaining the script within our existing setup by keeping Line 16 and Line 17 in .eslintrc.json.

package.json
Mdemenko1: I think it should be a separate ticket(issue) created for further work. I adjusted the code according to the PR comment

ivonne-hernandez (approved):

PR #927, opened 2024-02-07
Feat/new benefit flag

ivonne-hernandez:
- Add a `New Benefit` flag to the `ProgramCard`.

- I refactored the styles for the `Can't find what you need` text by removing the `important` value and adding `display:flex`.
- I moved the `result-program` styles into their own file.

Desktop view of the `New Benefit` flag:
Screenshot 2024-02-07 at 9 37 49 AM

Mobile view of the `New Benefit` flag:
Screenshot 2024-02-07 at 9 37 35 AM


src/Components/Results/211Button/211Button.tsx
```diff
@@ -5,7 +5,7 @@ const HelpButton = () => {
   return (
     <div className="help-text-for-211-button">
       <h2 className="text-center help-text-for-211-button-font">{`Can't find what you need?`}</h2>
-      <div>
+      <button className="hover-button">
```
CalebPena, line 8: This might be fixed by pulling from `feat/result_redesign`, but can you make this back into a `div`

CalebPena (approved):
This is great. I just have one question about changing the 211 button back to a button.

sydney-devine (changes requested):
Let's make the font/flag slightly bigger. Worried it won't be super legible as is.

myang1010 (approved):
Looks good! It's more logical for ProgramCard to have its own CSS class.

sydney-devine (approved):

PR #951, opened 2024-02-15
Refactor/results padding

ivonne-hernandez:
-  Reduce padding between tab and Filter
-  Calculations on headings and benefits cards need to be per month. They're displaying annual value right now. 
- Program page: move estimated values box to the right of the name of benefits box
- Program page: Make get help applying box documents box wider on desktop + paragraph wider
- Program page desktop: smallest font should be 16, go up from there for rest of headings.
- Padding under 2-1-1 button (between button & footer)
- Remove the 211 text and button from the 211 site.

- Some of the links were not working, we'll definitely need to test these to make sure they're fixed now.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,7 @@
   "plugins": ["react"],
   "rules": {
     "react/react-in-jsx-scope": "off",
-    "react/prop-types": "off"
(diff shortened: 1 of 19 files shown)


CalebPena (commented):
Looks good, could you just merge with the base branch.

PR #969, opened 2024-02-27
Combine no insurance and dont know options and add VA tile to insurance options

myang1010:
- Combined no insurance and don't know into "I don't have or know if I have health insurance"
- Added `va` to insurance options

- Removed the "dont_know" checks in validationFunctions and confirmation page

I was thinking to leave "dont_know" alone at first. But after going through the whole process, it makes sense to remove "dont_know" and "va" in a few places. Please let me know if there's any concerns. Thanks.


CalebPena (approved):
LGTM

Issue #801, opened 2023-12-08
As an administrator, I would like the ability to display a warning message on a benefit (example: if there is a localized delay or backlog within a county)


sydney-devine:
Icon should be middle aligned with the text in the box. 
- Make icon red
- background: ECDEED
- 1 px border: #978398
- ignore the linked text in the mockup. It's going to just be plain text. 

**Icon**
![Warning](https://github.com/Gary-Community-Ventures/benefits-calculator/assets/109622199/6c6826ee-bb82-4b5a-bc57-06920998961f)

![image](https://github.com/Gary-Community-Ventures/benefits-calculator/assets/109622199/85ed0671-f72d-4532-9cb6-d4b9bbba3278)

CalebPena:
Added the `warning` to program type. https://github.com/Gary-Community-Ventures/benefits-calculator/pull/902

CalebPena:
Similar component in https://github.com/Gary-Community-Ventures/benefits-calculator/blob/main/src/Components/BackToScreen/BackToScreen.tsx

sydney-devine:
@CalebPena add instructions on how to test this pls.

CalebPena:
1. Go to http://localhost:8000/api/translations/admin/programs/
2. Click edit on one of the programs that you want to add a warning.
3. Click on the link that says "Warning"
4. Add warning message

FYI empty warnings will be an empty string.

PR #989, opened 2024-03-11
801/warning message

Mdemenko1:

Work for the issue #801 

Show warning message if program has it

### Desktop view:

image

### Mobile view: 

image

### DIFF
--- a/src/Components/Results/ProgramPage/ProgramPage.tsx
+++ b/src/Components/Results/ProgramPage/ProgramPage.tsx
@@ -6,6 +6,8 @@ import BackAndSaveButtons from '../BackAndSaveButtons/BackAndSaveButtons.tsx';
 import { FormattedMessage } from 'react-intl';
 import { formatToUSD } from '../Results.tsx';
 import './ProgramPage.css';
+import BackToScreen from '../../BackToScreen/BackToScreen.tsx';
+import WarningMessage from '../../WarningComponent/WarningMessage.tsx';
 
 type ProgramPageProps = {
   program: Program;
@@ -80,6 +82,7 @@ const ProgramPage = ({ program }: ProgramPageProps) => {
         {displayIconAndHeader(program)}
         {displayEstimatedValueAndTime(program)}
       </div>
+      {program.warning.default_message && <WarningMessage message={program.warning.default_message} />}
       <div className="apply-online-button">
         <a href={program.apply_button_link.default_message} target="_blank">
           <FormattedMessage id="results.apply-online" defaultMessage="Apply Online" />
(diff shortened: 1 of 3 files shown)


src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -6,6 +6,8 @@ import BackAndSaveButtons from '../BackAndSaveButtons/BackAndSaveButtons.tsx';
 import { FormattedMessage } from 'react-intl';
 import { formatToUSD } from '../Results.tsx';
 import './ProgramPage.css';
+import BackToScreen from '../../BackToScreen/BackToScreen.tsx';
```
myang1010, line 9: Didn't see where we get to use this import. Remove it if it's not necessary.

src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -80,6 +82,7 @@ const ProgramPage = ({ program }: ProgramPageProps) => {
         {displayIconAndHeader(program)}
         {displayEstimatedValueAndTime(program)}
       </div>
+      {program.warning.default_message && <WarningMessage message={program.warning.default_message} />}
```
myang1010, line 85: Here we probably want to pass in warning in Translation type, so it can be translated.

src/Components/WarningComponent/WarningMessage.tsx
```diff
@@ -0,0 +1,20 @@
...
+type WarningMessageProps = {
+  message: string;
+};
+
+const WarningMessage = ({ message }: WarningMessageProps) => {
+  return (
+    <div className="warning-message">
+      <WarningIcon className="warning-icon" />
+      <p>
+        <FormattedMessage id="Warning.message" defaultMessage={message} />
```
myang1010, line 14: Here we can use `ResultsTranslate`.

myang1010 (commented):
Hi Marianna, I've made some suggestions below. Also, the warning on the desktop view appears slightly wider to me. I think it would be better to narrow it down to match the width of the program detail section below. Thanks.

myang1010 (approved):
Looks good.

src/Components/WarningComponent/WarningMessage.css
```diff
@@ -0,0 +1,20 @@
...
+  width: 75%;
+}
+
+.warning-icon {
+  min-width: 2rem;
+  min-height: 2rem;
+  width: 2rem;
+  height: 2rem;
+  filter: invert(33%) sepia(96%) saturate(7216%) hue-rotate(355deg) brightness(92%) contrast(120%);
+}
```
CalebPena, line 20:
```suggestion
}

.warning-icon path {
  fill: #ff0000;
}
```

CalebPena (approved):
Looks good. Added one improvement that I think you could make.

PR #991, opened 2024-03-13
Refactor frontend config to backend

mexi-cano:
-  Moved components off of frontend and into the new django admin regional config backend 
- Set two types to 'any' for sake of completing ticket.

Raw API response preview:
Screenshot 2024-03-13 at 2 50 27 PM


Mdemenko1:
@mexi-cano Could you leave instructions what needs to be changed on the back end so it is possible to run it? Thank you

mexi-cano:
> @mexi-cano Could you leave instructions what needs to be changed on the back end so it is possible to run it? Thank you

Absolutely. I've attached a text file with all the configs that need to be added to the backend under `Config`. @msrezaie 

[configs.zip](https://github.com/Gary-Community-Ventures/benefits-calculator/files/14637976/configs.zip)

msrezaie:
> Could you leave instructions what needs to be changed on the back end so it is possible to run it?


Just need to save each of the JSON instances from the sent zip file as a Configuration object in the backend through django admin. For example:
image

After saving all of them:
image

src/Components/Confirmation/Confirmation.js
msrezaie, line 1: The same issue with the structure of the config coming from the backend occurs in step-11 (confirmation screen) as well, where the Immediate needs section remains empty despite selecting some of the options:

image

I was able to make the messages show up with this quick fix

`Line 534: {relatedOptionsList[option].message ?? relatedOptionsList[option].props.defaultMessage}{' '}`

msrezaie (commented):
The config

src/Assets/validationFunctions.tsx
```diff
@@ -51,20 +58,24 @@ const displayAgeHelperText: MessageFunction<string> = (applicantAge) => {
...
   const numberMustBeFiveDigitsLongRegex = /^\d{5}$/;
   if (numberMustBeFiveDigitsLongRegex.test(zipcode.toString())) {
     //this means that the zipcode input passed the regex test so we can just return false since there is no error
     //this additional test checks the zipcode input against all CO zipcodes
-    return !Object.keys(countiesByZipcode).includes(zipcode.toString());
+    if (countiesByZipcode) {
+      return !Object.keys(countiesByZipcode).includes(zipcode.toString());
+    }
+    // If countiesByZipcode is undefined, return false (assuming no error)
+    return true;
```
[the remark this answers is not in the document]
mexi-cano: This is needed otherwise function returns undefined if `!countiesByZipcode`

src/Components/HouseholdDataBlock/HouseholdDataBlock.js
```diff
@@ -188,9 +188,12 @@ const HouseholdDataBlock = ({ handleHouseholdDataSubmit }) => {
...
   const createMembersAdded = (member, index) => {
     let relationship = relationshipOptions[member.relationshipToHH];
-    if (relationship === undefined) {
-      relationship = <FormattedMessage id="relationshipOptions.yourself" defaultMessage="Yourself" />;
-    }
+    relationship =
+      relationship === undefined ? (
+        <FormattedMessage id="relationshipOptions.yourself" defaultMessage="Yourself" />
+      ) : (
+        relationship.message
```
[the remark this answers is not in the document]
mexi-cano: Great catch. updated.

src/Components/AccordionsContainer/AccordionsContainer.tsx
```diff
@@ -11,8 +13,36 @@ type Props = {
...
+    };
+  });
+};
+
 const AccordionsContainer = ({ componentDetails, submitted }: Props) => {
-  const { formData } = useContext(Context);
+  const { formData, config } = useContext(Context);
+  let formattedBenefits: BenefitAccordion[] = [];
+
+  if (config) {
```
CalebPena, line 40: There seems to be a lot of checking if config is defined. Because of this, I think that we should create a config hook to check if the config is defined.

For example:
```ts
function useConfig(name: string) {
  const { config } = useContext(Context);

  if (config === undefined) {
    throw new Error('config is not defined');
  }
  
  if (config[name] === undefined) {
    throw new Error(`'${name}' does not exist in the config`);
  }

  return config[name];
}
```

Usage
```ts
const Component = () => {
  // Should be type safe
  const zipcodes = useConfig('zipcodes');

  ...
}
```

src/Components/Config/configHook.tsx
```diff
@@ -4,22 +4,142 @@ import { ConfigApiResponse } from '../../Types/Config';
...
+  };
+
+  if (typeof item !== 'object' || item === null) return item;
 
   if (item.hasOwnProperty('_label') && item.hasOwnProperty('_default_message')) {
     const { _label, _default_message } = item as Item;
     return <FormattedMessage id={_label} defaultMessage={_default_message} />;
   }
 
+  if (item.hasOwnProperty('icon') && item.hasOwnProperty('message')) {
```
CalebPena, line 56: Can we move this logic to it's own function?

src/Components/Confirmation/Confirmation.js
```diff
@@ -530,7 +529,7 @@ const Confirmation = () => {
       return (
         <p key={option} className="bottom-margin">
           {' '}
-          {relatedOptionsList[option]}{' '}
+          {relatedOptionsList[option].message ?? relatedOptionsList[option].props.defaultMessage}{' '}
```
CalebPena, line 532: I am a little confused how this one works. It looks like you are checking if `relatedOptionsList[option].message` is `undefined` and it is, then you are using `relatedOptionsList[option].props.defaultMessage`. Is the problem that `relatedOptionsList[option]` a `FormattedMessage`? If so, then does this work?
```suggestion
          {relatedOptionsList[option]}{' '}
```

CalebPena (changes requested):
I see that you wrapped all of the `{"_label": "label", "_default_message": "message"}` in an object with one message key. What do you think about removing the "message" object?

image

mexi-cano:
> > @mexi-cano Could you leave instructions what needs to be changed on the back end so it is possible to run it? Thank you
> 
> Absolutely. I've attached a text file with all the configs that need to be added to the backend under `Config`. @msrezaie
> 
> [configs.zip](https://github.com/Gary-Community-Ventures/benefits-calculator/files/14637976/configs.zip)

Uploaded new backend config structures.
[configs_v2.zip](https://github.com/Gary-Community-Ventures/benefits-calculator/files/14791070/configs_v2.zip)

CalebPena (changes requested):

src/Assets/validationFunctions.tsx
```diff
@@ -22,13 +26,13 @@ function useErrorController(hasErrorFunc: ValidationFunction<any>, messageFunc:
     setSubmittedCount(submittedCount + 1);
   };
 
-  const updateError: ValidationFunction<any> = (value, formData, countiesByZipcode) => {
+  const updateError: ValidationFunction<any> = (value, formData) => {
     const updatedHasError = hasErrorFunc(value, formData, countiesByZipcode);
```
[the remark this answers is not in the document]
mexi-cano: Updated.

CalebPena (approved):
Looks good.

CalebPena:
I am not going to merge yet, because this will break everyone's local environment. I created an [issue](https://github.com/Gary-Community-Ventures/benefits-api/issues/316) to add a script that will add this, so people only have to run one command to set up their local. I think that we should complete that issue before we merge this.

CalebPena:
@mexi-cano Can you fix the merge conflicts so that I can merge.

Issue #971, opened 2024-02-28
Add help bubbles throughout rest of screener


sydney-devine:
Put following into help bubbles (see income question for visual)

- Step 4. "`This is usually family members whom you live with and share important resources with like food and bills.`" 
- Income question on household members block, if they select yes: "`Answer the best you can. You will be able to include additional types of income. The more you include, the more accurate your results will be.`" (this is currently helper text under "What type of income have you had most recently?" 
- Step 6 "`Add up expenses for everyone who lives in your home. This includes costs like child care, child support, rent, medical expenses, heating bills, and more. We will ask only about expenses that may affect benefits. We will not ask about expenses such as food since grocery bills do not affect benefits.`"
- Step 7. Rewrite question stem to say "How much does your whole household have right now in cash, checking or savings accounts, stocks, bonds, or mutual funds?" Put "`In some cases, eligibility for benefits may be affected if your household owns other valuable assets such as a car or life insurance policy.`" into help bubble. 
- Step 8. "`This information will help make sure we don't give you results for benefits you already have.`"

kbottagaro:
@CalebPena The question should still read, "How much does your whole household have right now in cash, checking or savings accounts, stocks, bonds, or mutual funds?" The sentence that starts, "In some cases . . . " is what goes in the help text. Otherwise looks good.

PR #999, opened 2024-03-21
971/help bubble text

Mdemenko1:

Work for the issue #971 
Adding help text behind help bubble

https://github.com/Gary-Community-Ventures/benefits-calculator/assets/65931890/e4b98f1c-453b-4566-b931-308817175adb

For question 7, the translation file needs to be updated to display correct help text `'In some cases, eligibility for benefits may be affected if your household owns other valuable assets such as a car or life insurance policy.'` and question stem to say `"How much does your whole household have right now in cash, checking or savings accounts, stocks, bonds, or mutual funds?"`


msrezaie (approved):
Looks good to me,

CalebPena (approved):
Is there a situation where we would set `isVisible` to `false`? If not, I think that we should remove it. Besides that, this looks good.

Issue #1001, opened 2024-03-21
Change language in card in step 9 that relates to child development


kbottagaro:
Related to this ticket - https://app.zenhub.com/workspaces/myfriendben-652eb93b7dc40e0935103f4c/issues/gh/gary-community-ventures/benefits-calculator/1000.

Change card in step 9 that reads "Concern about your baby or toddler's development" to instead read "Concern about your child's development."

kbottagaro:
Sorry - I'm not sure why the link isn't working. But yes, this ticket
relates to api#310. The idea is that we're changing baby/toddler to child
and, in conjunction with that, adding Child First as an immediate need.

rickyctd:
Hi @kbottagaro is what is circled in the attached screenshot the only location where this text change needs to occur as requested in this ticket? Thanks so much! 

![image](https://github.com/Gary-Community-Ventures/benefits-calculator/assets/130072805/f9fc32f5-7ed1-4e44-9c41-c8f0c2a5e38c)

kbottagaro:
Yes, that's it. Thank you!

PR #1009, opened 2024-04-04
Changing language step 9

Mdemenko1:

Changing language for the tile in the step 9 for the issue #1001 

- [ ] Needs to be changed in translations as well


CalebPena (approved):
LGTM

PR #1010, opened 2024-04-04
Fix Dollar Sign

CalebPena:
- Add dollar sign back into inputs.


myang1010 (approved):
Looks good!

src/Components/Textfield/Textfield.js
```diff
@@ -36,15 +36,14 @@ const Textfield = ({ componentDetails, data, handleTextfieldChange, index, submi
       helperText={errorController.showError && errorController.message(data[inputName])}
       inputProps={numericField ? { pattern: '[0-9]* .', inputMode: 'decimal' } : {}}
       InputProps={
-        (dollarField
```
myang1010, line 39: If we add `...` in the front, we don't need to pass in `sx: { backgroundColor: '#FFFFFF' }` twice.

ivonne-hernandez (approved):

Issue #977, opened 2024-03-04
Add "twice a month" option to dropdown menu for question "How often are you paid this income"


kbottagaro:
Add "twice a month" option to dropdown menu for question "How often are you paid this income"

See related ticket for help text - https://app.zenhub.com/workspaces/myfriendben-652eb93b7dc40e0935103f4c/issues/gh/gary-community-ventures/benefits-calculator/978.

NOTE: Someone who gets paid twice a month on the same dates each month gets 24 paychecks over 12 months. Someone who gets paid every other week gets 26 paychecks over 12 months.

kbottagaro:
@CalebPena I see this in the current sprint and thought it was going to be part of today's releases. Can we include this in the April 12 releases? This is important to Salud.

rickyctd:
@kbottagaro @CalebPena to clarify what needs to be done for this ticket, is it to solely add the "twice a month" option on the drop down? The original ticket text links to 978 which doesn't appear to have been done either. The original text on this ticket also includes a "note" and it is unclear if that is meant to be a note internally to clarify for us or if it needed to be displayed. If it needs to be displayed, it does not indicate how you'd like that displayed.

rickyctd:
@Mdemenko1 I just clarified with @CalebPena and this ticket is just to add the option on the drop down and not adding the note or help text on display. That will be the separate ticket for next week.

kbottagaro:
Hi, Ricky. Thanks for asking for clarification. Ticket 977 involves adding
the "twice a month" option to the income frequency dropdown menu. I assume
that this change involves updating the backend calculations to reflect this
method of income reporting. As a nontech person, I had thought the updated
calculation would be implied in the ticket - but please let me know if you
need a separate ticket for that.

And yes, ticket 978 is related. The help text would be displayed like the
other help text bubbles we added last week with question marks.

Please let me know if you need further clarification. So appreciate your
help!

kbottagaro:
Seems like our messages crossed. I understand from Caleb the calculations
are updated on the backend because MFB had twice a month when it was first
created.

I also understand the help text will be a separate ticket for next week.

rickyctd:
@kbottagaro Thanks so much for the context and clarification!

PR #1012, opened 2024-04-09
977: dropdown option add

Mdemenko1:

Work for the issue #977. Adding an option to the dropdown in the question 5

image

### DIFF
--- a/src/Assets/frequencyOptions.tsx
+++ b/src/Assets/frequencyOptions.tsx
@@ -3,7 +3,7 @@ import { FormattedMessage } from 'react-intl';
 const frequencyOptions = {
   weekly: <FormattedMessage id="frequencyOptions.weekly" defaultMessage="every week" />,
   biweekly: <FormattedMessage id="frequencyOptions.biweekly" defaultMessage="every 2 weeks" />,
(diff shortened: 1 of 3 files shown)


src/Assets/frequencyOptions.tsx
```diff
@@ -3,6 +3,7 @@ import { FormattedMessage } from 'react-intl';
 const frequencyOptions = {
   weekly: <FormattedMessage id="frequencyOptions.weekly" defaultMessage="every week" />,
   biweekly: <FormattedMessage id="frequencyOptions.biweekly" defaultMessage="every 2 weeks" />,
+  twiceMonth: <FormattedMessage id="frequencyOptions.twiceMonth" defaultMessage="twice a month" />,
```
msrezaie, line 6: The twice a month option in the backend is named `semimonthly`. Please name this the same in here as well.

msrezaie (changes requested):
On the 'confirmation-page' if twice a month option is selected, the income is returned zero for annual like so:
image

msrezaie (approved):

CalebPena (approved):

PR #1088, opened 2024-05-21
Feat/aria labels

ivonne-hernandez:
- I replaced the current `aria-labels` with translated `aria-labels`.


src/Components/EmailResults/EmailResults.tsx
```diff
@@ -49,6 +50,10 @@ const EmailResults = forwardRef(function EmailResults(
   });
   const [emailSubmitted, setEmailSubmitted] = useState(false);
   const [phoneSubmitted, setPhoneSubmitted] = useState(false);
+  const closeAriaLabelProps = {
+    id: 'emailResults.close-AL',
+    defaultMsg: 'goodbye',
```
CalebPena, line 55: I don't think that this is a good aria-label for what the button does.

src/Components/Results/Translate/TranslateAriaLabel.tsx
CalebPena, line 1: You are not using this as a component, so it should not be a component. I would recommend using the `useIntl` hook in all of the components with `aria-label`. Another option is to wrap `useIntl` with a custom hook. For example:

```tsx
function useAriaLabel(id: string, defaultMsg: string) {
  const intl = useIntl();

  return intl.formatMessage({ id: id, defaultMessage: defaultMsg });
};
```
It would then be used like:
```tsx
const label = useAriaLabel('id', 'message');

return <button aria-label={label}>Open</button>
```

src/Components/Share/Share.js
```diff
@@ -73,9 +74,14 @@ const Share = forwardRef(function Share({ close }, ref) {
     };
   };
 
+  const closeALProps = {
+    id: 'share.close',
```
CalebPena, line 78: Can we add `.ariaLabel` to the end of the labels to prevent potential translation label conflicts.

src/Components/TwoOneOneComponents/TwoOneOneHeader/TwoOneOneHeader.js
```diff
@@ -126,7 +128,7 @@ const TwoOneOneHeader = ({ handleTextfieldChange }) => {
         sx={{
           zIndex: 1000,
           [`& .MuiDrawer-paper`]: {
-            mt: '50px',
+            mt: '75px',
```
CalebPena, line 131: Can we make this `rem`

CalebPena (changes requested):

src/Components/TwoOneOneComponents/TwoOneOneHeader/TwoOneOneHeader.js
ivonne-hernandez: Since this is a one off I think it's fine to leave it as is.

PR #1095, opened 2024-05-22
Fix New Benefit Flag Wrap in Other Languages

myang1010:
- Use useRef to get the width of "new benefit" flag, and apply that value to the left offset of "low confidence" flag
- Updated "new-program-flag" and "low-confidence-flag" css as they both tend to wrap when the text gets too long

- Before
329005530-ce7a9bb9-40a7-47a2-8f9a-8b4016e3eafa

- After in different languages
> c

> a

> b

### DIFF
--- a/src/Components/Results/Programs/ProgramCard.tsx
+++ b/src/Components/Results/Programs/ProgramCard.tsx
@@ -1,9 +1,11 @@
 import { Link, useParams } from 'react-router-dom';
 import { Program } from '../../../Types/Results';
 import { FormattedMessage } from 'react-intl';
-import { formatMonthlyValue, programValue } from '../FormattedValue';
+import { formatMonthlyValue } from '../FormattedValue';
 import ResultsTranslate from '../Translate/Translate';
 import { useEffect, useRef, useState } from 'react';
+import { useContext } from 'react';
+import { Context } from '../../Wrapper/Wrapper';
 import './ProgramCard.css';
 
 type ProgramCardProps = {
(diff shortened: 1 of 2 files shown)


src/Components/Results/Programs/ProgramCard.tsx
```diff
@@ -44,12 +58,16 @@ const ProgramCard = ({ program }: ProgramCardProps) => {
   return (
     <div className="result-program-container">
       {program.new && (
```
CalebPena, line 60: I think that is might have been easier if you created a flags container that is positioned absolutely at the top left, that you can than put the flags in it. That way you don't need to calculate the offset.

src/Components/Results/Programs/ProgramCard.tsx
```diff
@@ -34,6 +34,20 @@ const ProgramCard = ({ program }: ProgramCardProps) => {
...
   const isMobile = size < 769 ? true : false;
 
+  const newProgramFlagRef = useRef<HTMLDivElement | null>(null);
+  const [leftOffset, setLeftOffset] = useState('0');
+  const [newProgramFlagWidth, setNewProgramFlagWidth] = useState<number>(0);
+
+  useEffect(() => {
+    if (program.new && newProgramFlagRef.current) {
+      const width = newProgramFlagRef.current.offsetWidth;
+      const remWidth = width / 16;
```
CalebPena, line 44: Can we add font size as a number to the style controller, and then divide by that. This breaks if we change the root font size.

CalebPena: Also, `theme.fontSize` will need to be added to the `useEffect` dependencies.

CalebPena (changes requested):
Just need to change the font size divisor to be dynamic.

src/Components/Results/Programs/ProgramCard.tsx
myang1010: Thanks for pointing this out. After looking at the style controller realized the `font-size` is already part of `css-variables`. So I just use that value instead of creating a new one.

src/Components/Results/Programs/ProgramCard.tsx
```diff
@@ -57,20 +59,22 @@ const ProgramCard = ({ program }: ProgramCardProps) => {
...
-        <div className="low-confidence-flag" style={{ left: program.new ? leftOffset : '0' }}>
-          <FormattedMessage id="results-low-confidence-flag" defaultMessage="Low Confidence" />
-        </div>
-      )}
+      <div className="result-program-flags-container">
+        {program.new && (
+          <div
+            className="new-program-flag"
+            ref={newProgramFlagRef}
+            style={{ width: newProgramFlagWidth > 0 ? `${newProgramFlagWidth}rem` : 'auto' }}
```
CalebPena, line 67: I think this can be removed.

src/Components/Results/Programs/ProgramCard.tsx
```diff
@@ -35,18 +37,18 @@ const ProgramCard = ({ program }: ProgramCardProps) => {
...
       const width = newProgramFlagRef.current.offsetWidth;
-      const remWidth = width / 16;
-      const paddedWidth = remWidth - 0.5;
+      const fontSize = parseFloat(theme.cssVariables['font-size']);
+      const remWidth = width / fontSize;
       setNewProgramFlagWidth(remWidth);
-      setLeftOffset(`${paddedWidth}rem`);
     }
-  }, [program.new]);
+  }, [program.new, theme.cssVariables['font-size']]);
```
CalebPena, line 51: I think that this can be removed.

CalebPena (changes requested):
I think that you can remove some code now that you have added a flag container.

CalebPena (approved):
LGTM

PR #1124, opened 2024-06-05
Feat/list co bens

ivonne-hernandez:
- I am implementing the `current-benefits` route and component which displays all of the current long and near term programs.

- I came across a persistent issue that broke the app entirely when using the program's `category` or `type`'s `default_message` string value. In particular some of the programs had categories/types that were not on my local repository list or had lowercase and not uppercase values that didn't match. It took a lot of time to troubleshoot this error making it the second time in which these fields break the entire app (the first time was when we changed the child care category name which broke the app). As such I recommend the following solution: adding a `categoryKeyName` field to all near and short term programs so that we're not obtaining the category/type value from a random string. This is a screenshot of the issue where we're ending up with two keys for the same category type and one category `Managing housing costs` for which I had nothing for on my local repository:
Screenshot 2024-06-04 at 4 46 52 PM

Desktop:
Screenshot 2024-06-05 at 11 21 17 AM

Mobile:
Screenshot 2024-06-05 at 10 57 14 AM


src/Components/CurrentBenefits/CurrentBenefits.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const [allNearTermPrograms, setAllNearTermPrograms] = useState<Program[]>([]);
+  const intl = useIntl();
+
+  useEffect(() => {
+    getAllLongTermPrograms().then((response) => {
+      setAllLongTermPrograms(response);
+    });
+
+    getAllNearTermPrograms().then((response) => {
+      setAllNearTermPrograms(response);
```
CalebPena, line 113: We can do this to prevent a lot of complexity in determining whether to use type or category in a lot of the functions. Additionally, there will only be a one line change when we change the API.
```suggestion
      response.map((program) => program.category = program.type)
      setAllNearTermPrograms(response);
```

src/Components/CurrentBenefits/CurrentBenefits.tsx
```diff
@@ -0,0 +1,226 @@
...
+    groupProgramsIntoCategories: (
+      programs: Program[],
+      typeOrCategoryField: 'type' | 'category',
+    ) => Record<CategoryName, Category>,
+  ) => {
+    const programsSortedByCategories = groupProgramsIntoCategories(programs, typeOrCategoryField);
+
+    const categoryHeaderIconAndPrograms = Object.keys(programsSortedByCategories).map((key: string, index) => {
+      const categoryName = key as CategoryName;
+      const category = programsSortedByCategories[categoryName];
```
CalebPena, line 167:
```suggestion
    const categoryHeaderIconAndPrograms = Object.entries(programsSortedByCategories).map(([key, category], index) => {
      const categoryName = key as CategoryName;
```

src/Components/CurrentBenefits/CurrentBenefits.tsx
```diff
@@ -0,0 +1,226 @@
...
+          <div className="category-heading-column">
+            <div
+              className="category-heading-icon"
+              aria-label={intl.formatMessage({
+                id: name.label,
+                defaultMessage: name.default_message,
+              })}
+              role="img"
+            >
+              <CategoryIcon />
```
CalebPena, line 182: Should the `role` and `aria-label` be on the category icon?

src/Components/CurrentBenefits/CurrentBenefits.css
```diff
@@ -0,0 +1,44 @@
+.co-benefits-container {
+  margin: 1rem;
```
CalebPena, line 2: Do we want to add a max width to this to be consistent with the rest of the app?

CalebPena (changes requested):
Can we remove all of the Colorado references in variables and class names? This is intended to be state independent.

CalebPena (approved):

PR #1136, opened 2024-06-11
Feat/use category key name

ivonne-hernandez:
- I implemented a way to have the program `category` and `type` fields be normalized and return the correct string value or throw an error if it is `undefined`. This will help us resolve issues that have come up whenever the `category` or `type`'s `defaultMessage` string is incompatible with what we have hardcoded in our repo for category/type values.

- As I tested out this PR by changing one of the program's `category.name` in admin and refreshing the page I discovered that the real problem lies in the `translations` app since it still uses the translation `id` value to obtain the string value (which may be incorrectly typed in admin) that we see on the front end and only uses the `default_message` field (which is what I changed on this PR) as a fallback:

Screenshot 2024-06-12 at 11 04 36 AM

### DIFF
--- a/src/cleanAPICategoryTranslation.ts
+++ b/src/cleanAPICategoryTranslation.ts
@@ -35,8 +35,8 @@ const alternateProgramTemplateCategories = [
 ];
 
 const normalizeDefaultMessage = (defaultMessage: string): string => {
-  const whiteSpaceCommasApostropheOrHyphenMinusRegex = /\s|,|'|-/g;
-  return defaultMessage.toLowerCase().replaceAll(whiteSpaceCommasApostropheOrHyphenMinusRegex, '');
+  const whiteSpaceCommasApostropheHyphenMinusNumbersRegex = /\s|,|'|-|[0-9]/g;
+  return defaultMessage.toLowerCase().replaceAll(whiteSpaceCommasApostropheHyphenMinusNumbersRegex, '');
 };
 
 const findClosestMatchingDefaultMessage = (defaultMessage: string): string => {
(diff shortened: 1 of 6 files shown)


src/cleanAPICategoryTranslation.ts
```diff
@@ -0,0 +1,77 @@
...
+    correctValue: 'Managing housing costs',
+  },
+  {
+    previousNormalizedValues: ['jobtraining'],
+    correctValue: 'Job resources',
+  },
+];
+
+const normalizeDefaultMessage = (defaultMessage: string): string => {
+  const whiteSpaceCommasApostropheOrHyphenMinusRegex = /\s|,|'|-/g;
```
CalebPena, line 38: What do you think about removing any non-character or digit? That way we can cover more edge cases.
```suggestion
  const whiteSpaceCommasApostropheOrHyphenMinusRegex = /[^a-zA-Z0-9]/g;
```

src/cleanAPICategoryTranslation.ts
```diff
@@ -0,0 +1,77 @@
...
+    // or does it have typos not related to whitespace, comma, apostrophe, or hyphen characters?
+    const maybeClosestMatchingCategory = alternateProgramTemplateCategories.find((programTemplate) =>
+      programTemplate.previousNormalizedValues.includes(normalizedDefaultMessage),
+    );
+
+    if (!maybeClosestMatchingCategory) {
+      throw new Error(`CategoryName is undefined`);
+    }
+    return maybeClosestMatchingCategory.correctValue;
+  }
```
CalebPena, line 67:
```suggestion
  }
  // is the default message invalid because it uses different wordings
  // or does it have typos not related to whitespace, comma, apostrophe, or hyphen characters?
  const maybeClosestMatchingCategory = alternateProgramTemplateCategories.find((programTemplate) =>
    programTemplate.previousNormalizedValues.includes(normalizedDefaultMessage),
  );

  if (!maybeClosestMatchingCategory) {
    throw new Error(`CategoryName is undefined`);
  }
  return maybeClosestMatchingCategory.correctValue;
```

CalebPena (approved):

src/cleanAPICategoryTranslation.ts
ivonne-hernandez: These are logically equivalent statements so it's just a matter of preference.

src/cleanAPICategoryTranslation.ts
ivonne-hernandez: Good call! I've added that case to the regex pattern and updated the comments as necessary.

CalebPena (approved):

PR #1201, opened 2024-07-25
Refactor/211 a11y menu

ivonne-hernandez:
- I refactored the `displayHamburgerMenu` function and rendering so that the header is still accessible when that component is open. 
- I fixed the spacing issue that was taking place with other languages on the header's desktop links.

Screenshot 2024-07-25 at 12 54 10 PM

Screenshot 2024-07-25 at 11 14 51 AM
Screenshot 2024-07-25 at 11 14 41 AM


CalebPena (approved):
Looks good. Just want to confirm that we do not want the drawer to cover the whole part below the header anymore.

ivonne-hernandez:
> Looks good. Just want to confirm that we do not want the drawer to cover the whole part below the header anymore.

I just added `height: 100vh` so that it would take up the rest of the screen height.

PR #1234, opened 2024-08-08
Split the program name and desc for styling purposes

ivonne-hernandez:
- I split the program name and description for styling purposes.

- [ ] I noticed that an `ErrorMessage` component does not render when the user selects `Yes` but then doesn't check any of the boxes.

- [ ] Please go through and make sure that nothing breaks. 
- [ ] I had to refactor the `Confirmation` component to account for the `category_benefits` schema. Please note that once this is approved, I'll have to go in and manually update all of these FM values.

Updated text and styling on `step-8`:
Screenshot 2024-08-08 at 4 12 59 PM

Updated `Confirmation` component:
Screenshot 2024-08-28 at 4 52 59 PM


src/Components/CheckboxGroup/BasicCheckboxGroup.js
```diff
@@ -27,13 +27,25 @@ const BasicCheckboxGroup = ({ stateVariable, options }) => {
     const optionKeys = Object.keys(options);
 
     const formControlLabels = optionKeys.map((optionKey) => {
+      const splitLabel = options[optionKey].props.defaultMessage.split(':');
```
CalebPena, line 30: Does this work for Spanish? You might have to split them into 2 separate messages in the config.

CalebPena (changes requested):

src/Components/Confirmation/Confirmation.js
```diff
@@ -564,17 +564,33 @@ const Confirmation = () => {
...
-        <p key={option} className="bottom-margin">
-          {' '}
-          {relatedOptionsList[option]}{' '}
-        </p>
-      );
-    });
+  const listAllTruthyValues = (selectedOptions, relatedOptionsList, stateVariableName) => {
+    if (stateVariableName === 'benefits') {
+      const mappedListItems = selectedOptions.map((option) => {
+        const { id: nameId, defaultMessage: nameDefaultMsg } = relatedOptionsList[option].name.props;
```
CalebPena, line 570: You can use `relatedOptionsList[option].name` as a formatted message.

src/Components/Confirmation/Confirmation.js
```diff
@@ -564,17 +564,33 @@ const Confirmation = () => {
...
+      });
 
-    return mappedListItems;
+      return mappedListItems;
+    } else {
+      const mappedListItems = selectedOptions.map((option) => {
+        return (
+          <p key={option} className="bottom-margin">
+            {' '}
+            {relatedOptionsList[option]}{' '}
```
CalebPena, line 587: What do these spaces do?

CalebPena (changes requested):

src/Components/Confirmation/Confirmation.js
ivonne-hernandez: They're extraneous so I removed them.

src/Components/Confirmation/Confirmation.js
ivonne-hernandez: I made this change.

CalebPena (approved):

PR #1240, opened 2024-08-13
Refactor/back button

ivonne-hernandez:
- I created a `PrevAndContinueButtons` component to reduce prop drilling and improve validation. This component now handles validation on a per-question basis using react-hook-form's `handleSubmit` function. By doing this, validation logic is encapsulated within each question's `form` component, making the code cleaner and easier to maintain.

- I replaced the hardcoded previous and continue buttons in the zipcode question with the new `PrevAndContinueButtons` component. This change helps standardize the navigation buttons across components, making the code more consistent and easier to manage.


src/Components/Steps/ZipcodeStep.tsx
```diff
@@ -141,7 +138,9 @@ export const ZipcodeStep = () => {
         />
         {shouldShowCountyInput && (
           <div>
-            <QuestionQuestion>{matchingQuestion.followUpQuestions[0].question}</QuestionQuestion>
+            {matchingQuestion.followUpQuestions && (
```
CalebPena, line 141: There is always going to be a follow up question, so you don't need this conditional.

src/Components/PrevAndContinueButtons/PrevAndContinueButtons.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { FormattedMessage } from 'react-intl';
+
+type PrevAndContinueButtonsProps = {
+  currentStepId: number;
+};
+
+const PrevAndContinueButtons = ({ currentStepId }: PrevAndContinueButtonsProps) => {
+  const { uuid } = useParams();
+  const navigate = useNavigate();
+  const backNavigationFunction = () => navigate(`/${uuid}/step-${currentStepId - 1}`);
```
CalebPena, line 13: Can we pass this function into this component so that if we need to do something different (like the household member question), we can have a different previous function?

CalebPena (approved):
Just one suggestion to prepare for a use case in the future.

src/Components/Steps/ZipcodeStep.tsx
ivonne-hernandez: Most questions don't have `followUpQuestions`. This also takes care of the `typescript` error.

PR #1376, opened 2024-11-15
White label routing

CalebPena:
- Add the state/white label to the url.
- Add a select your state question (we skip for now because we only have one state) (set `REACT_APP_STATE` to `CO` to navigate to the co white label).

- Refactor the referrer question.
- Update all of the links to go to the right url.

- We need to have a default config so that we can have a header/footer on the non white label select language page. We need to add `footer_data`, `referrer_data`, `language_options`, and `feedback_links` under a white label with a `_default` code.

**The flow of the first steps looks now.**
image

Most links should point to a specific white label. e.x. `/co/step-1/`


src/Types/WrapperContext.ts
```diff
@@ -15,8 +15,8 @@ export interface WrapperContext {
   setTheme: React.Dispatch<React.SetStateAction<'default' | 'twoOneOne'>>;
   styleOverride: any;
   pageIsLoading: boolean;
-  screenDoneLoading: () => void;
-  getReferrer: (id: keyof ReferrerData) => any;
+  setScreenLoading: (loading: boolean) => void;
+  getReferrer: (id: keyof ReferrerData, defualt?: ReferrerDataValue) => any;
```
ivonne-hernandez, line 19: What do you think about fixing the `default` typo?

src/Components/Steps/Referrer.tsx
```diff
@@ -0,0 +1,186 @@
...
+      }
+
+      return { ...val, otherReferrer: '' };
+    });
+
+  type FormSchema = z.infer<typeof formSchema>;
+
+  const isOtherSource =
+    !(formData.referralSource === undefined || formData.referralSource === '') &&
+    !(formData.referralSource in referralOptions);
```
ivonne-hernandez, line 66: What do you think about not using !== ?

ivonne-hernandez (approved):

PR #1386, opened 2024-11-26
Change steps based on the config

CalebPena:
- Refactor the step function to be hooks that use the referrer data.
- Refactor the confirmation page so each box has its own component.
- Refactor the confirmation page to use the step directory.
- Change the `FormData` interfaces to be types.

_There are some config changes that were made to go with this pr._

_Don't forget to update all the translations. I updated some of the labels on the confirmation page._


ivonne-hernandez (changes requested):
We discovered your // TODO: update pageTitleTags.ts still needs to be implemented.

src/Assets/pageTitleTags.ts
```diff
@@ -1,32 +1,24 @@
...
+export const QUESTION_TITLES: Record<QuestionName, string> = {
+  zipcode: 'Zip and County',
+  householdSize: 'Number of HH Members',
+  householdData: 'Individual HH Member',
+  hasExpenses: 'Expenses',
+  householdAssets: 'Assets',
+  hasBenefits: 'Existing Benefits',
+  acuteHHConditions: 'Near Term Help',
+  referralSource: 'Referral',
+  signUpInfo: 'Optional Sign Up',
```
ivonne-hernandez, line 12: What do you think about having Sydney create a ticket to move these to the back end config and translate them?

ivonne-hernandez (approved):

src/Assets/pageTitleTags.ts
CalebPena: Can you create an issue to discuss it. I don't think that we need to move them to the backend because they should be the same for each state, but I think that we would want to translate them. Although, I don't know how that would affect analytics or SEO. At the very least we should not abbreviate "household"

ivonne-hernandez (approved):

PR #1388, opened 2024-12-03
Bugfix: Age being overridden on old screens

kbrey:

Fixes an issue where when a user returns to a screen from before age was replaced with birth year/month, the age displays as NaN, and when the user edits a non-household member question there's an error in the results. 
- Add back `age` to form data with deprecated comment for clarity
- Checks for undefined birth month/year and displays "N/A" // TODO: find a better translated default
- Update calcAge to default to `age` if `birthMonth` or `birthYear` are not defined

- Pulled calcAge into `age.tsx` in Assets, as it is used outside of the HouseholdDataBlock
- Created `getCurrentMonthYear` helper function, used in `calcAge` and `AgeInput`
- Created `useFormatBirthMonthYear` hook to share code that translates and formats the birth month and year into `{month}/{year}`


src/Assets/age.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export function useFormatBirthMonthYear() {
+  const translateNumber = useTranslateNumber()
+  return (birthMonth: number | null, birthYear: number | null) => {
+    if(birthMonth !== undefined && birthYear !== undefined) {
+      const formattedMonth = translateNumber(String(birthMonth).padStart(2, '0'));
+      const formattedYear = translateNumber(birthYear as number);
+      return `${formattedMonth}/${formattedYear}`;
+    }
+    return "N/A"
```
kbrey, line 41: @CalebPena is there a translated default that I can use?

CalebPena: If it does not exist, let's just not show it on the confirmation page.

CalebPena (approved):
Looks good. Can you pull from the dev branch. The confirmation cards got moved into a new file, so I think that you need to update that in the new file. The new file is called `/src/Components/Confirmation/ConfirmationSteps.tsx`. Let me know if you have any questions.

src/Assets/age.tsx
```diff
@@ -11,18 +12,24 @@ export function getCurrentMonthYear() {
...
+export function hasBirthMonthYear(member: MemberBirthMonthYear | Partial<HouseholdData>): member is MemberBirthMonthYear {
+  const {birthMonth, birthYear} = member;
   const hasBirthYear = birthYear !== undefined && birthYear !== null;
   const hasBirthMonth = birthMonth !== undefined && birthMonth !== null;
   return hasBirthYear && hasBirthMonth;
 }
 
-export function calcAge(age: number, birthYear: number, birthMonth: number) {
-  const hasBirthDate = hasBirthMonthYear(birthYear, birthMonth);
+export function calcAge(member: Partial<HouseholdData>) {
```
CalebPena, line 24: Can you just use `HouseholdData` here?

CalebPena (approved):
Looks good. I added one optional suggestion.

src/Assets/age.tsx
kbrey: @CalebPena Thanks for the suggestion, I cleaned up the types.

CalebPena (approved):

PR #1391, opened 2024-12-03
Trim input text

CalebPena:
- Trim the user entered text.

_I also removed the white space for the other referrer, as I can't think of a reason it would be useful to store the whitespace the user entered._

### DIFF
--- a/src/Components/Steps/ZipcodeStep.tsx
+++ b/src/Components/Steps/ZipcodeStep.tsx
@@ -8,13 +8,13 @@ import { FormattedMessageType } from '../../Types/Questions';
 import { FormData } from '../../Types/FormData';
 import { Context } from '../Wrapper/Wrapper';
 import { updateScreen } from '../../Assets/updateScreen';
-import ErrorMessageWrapper from '../ErrorMessage/ErrorMessageWrapper.tsx';
-import { useConfig } from '../Config/configHook.tsx';
+import ErrorMessageWrapper from '../ErrorMessage/ErrorMessageWrapper';
+import { useConfig } from '../Config/configHook';
 import * as z from 'zod';
 import QuestionHeader from '../QuestionComponents/QuestionHeader';
 import QuestionLeadText from '../QuestionComponents/QuestionLeadText';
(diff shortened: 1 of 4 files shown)


src/Components/Steps/ZipcodeStep.tsx
```diff
@@ -50,27 +61,19 @@ export const ZipcodeStep = () => {
   });
 
   const currentZipcodeValue = watch('zipcode');
-  const shouldShowCountyInput = zipcodeSchema.safeParse(currentZipcodeValue).success;
+  const parsedZipCode = zipcodeSchema.safeParse(currentZipcodeValue);
```
ivonne-hernandez, line 64: What do you think about using the original expression? It seems more readable and explicit in terms of why the parsed zip code is being used.

src/Components/Steps/ZipcodeStep.tsx
```diff
@@ -164,7 +167,7 @@ export const ZipcodeStep = () => {
                           id="questions.zipcode-a-disabledSelectMenuItemText"
                           defaultMessage="Select a county"
                         />,
-                        countiesByZipcode[watch('zipcode')],
+                        countiesByZipcode[parsedZipCode.data],
```
ivonne-hernandez, line 170: What do you think about using the original expression? I understand that there are performance trade offs with using `watch` in line, but it's more readable and explicit in terms of what the `zipcode` value is.

ivonne-hernandez (changes requested):
Please see the comments regarding some of the refactored `Zipcode` variables and expressions.

src/Components/Steps/ZipcodeStep.tsx
CalebPena: The new way, removes the spaces, so that we don't get a key error.

src/Components/Steps/ZipcodeStep.tsx
CalebPena: This fixes a TS error, and allows us to use the parsed zip code in multiple places.

ivonne-hernandez (approved):

PR #1409, opened 2024-12-19
Hidden citizenship filters

CalebPena:
- Only show some warning messages based on the citizenship filters

- Hide some citizenship filters and select them based on the form data.
- Refactor the citizenship filters on change function.

### DIFF
--- a/src/Components/Results/ProgramPage/ProgramPage.tsx
+++ b/src/Components/Results/ProgramPage/ProgramPage.tsx
@@ -114,6 +114,8 @@ const ProgramPage = ({ program }: ProgramPageProps) => {
   const warningMessages = useMemo(() => {
     return program.warning_messages.filter((warningMessage) => {
       if (warningMessage.legal_statuses.length === 0) {
+        // if no legal statuses are selected,
+        // then assume that the waring is for all legal statuses
         return true;
       }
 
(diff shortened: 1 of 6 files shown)


src/Components/Results/ProgramPage/ProgramPage.tsx
```diff
@@ -110,6 +111,22 @@ const ProgramPage = ({ program }: ProgramPageProps) => {
   };
   const value = useFormatYearlyValue(program);
 
+  const warningMessages = useMemo(() => {
+    return program.warning_messages.filter((warningMessage) => {
+      if (warningMessage.legal_statuses.length === 0) {
+        return true;
+      }
```
ivonne-hernandez, line 118: What do you think about adding a comment here to denote that this if statement handles a `warning_message` that is not tied to a legal status and therefore applies generally.

src/Components/Results/Programs/Filter.tsx
```diff
@@ -57,78 +37,82 @@ export const Filter = () => {
...
-              <Checkbox
-                checked={filtersChecked[citizenshipFCLKey]}
-                onChange={() => handleFilterSelect(citizenshipFCLKey)}
-              />
-            }
-            className="vertical-align"
-          />,
-        );
-      } else if (isSubfilterAndMainFilterIsChecked) {
+    // if all filters are unchecked set citzenship to false
```
ivonne-hernandez, line 68: `// if all filters are unchecked set citizenship to false`

ivonne-hernandez (approved):

PR #1420, opened 2025-01-07
Load config

CalebPena:
- Refactor how we load the config when we change the white label, so that the white label is always set before we load the config. This was previously causing white screens.


src/App.tsx
```diff
@@ -135,6 +135,7 @@ const App = () => {
     setFormData({ ...formData, [name]: value });
   };
 
+  // TODO: update updateScreen hook to not take in override uuid
```
ivonne-hernandez, line 138: Was this implemented?

ivonne-hernandez (approved):

src/App.tsx
CalebPena: This needs to be implemented when this function is removed.

PR #1449, opened 2025-01-23
Refactor/email results

ivonne-hernandez:
- I replaced the `EmailResults` component and css file with a `SaveMyResultsModal`, `EmailTextfield`, `PhoneTextfield` using react-hook-form and zod.


src/Components/SaveMyResultsModal/EmailTextfield.tsx
```diff
@@ -0,0 +1,111 @@
...
+      screen: uuid,
+      email: data.email,
+      type: 'emailScreen',
+    };
+
+    try {
+      await postMessage(messageBody);
+      setSnackbar({ open: true, message: snackbarMessage });
+    } catch (error) {
+      throw new Error(`${error}`);
```
CalebPena, line 68: Can we set a snackbar message if there is an error saying that could not send an email?

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+        for (const char of value) {
+          if (/\d/.test(char)) {
+            newString += char;
+          }
+        }
+
+        return newString;
+      })
+      .refine((value) => value.length === 10, {
```
CalebPena, line 47: Zod has a length validator. https://zod.dev/?id=strings#:~:text=z.string().length(5)%3B

Also you don't need the message because of the `errorMap`

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
```diff
@@ -0,0 +1,126 @@
...
+      screen: uuid,
+      phone: data.phone,
+      type: 'textScreen',
+    };
+
+    try {
+      await postMessage(messageBody);
+      setSnackbar({ open: true, message: snackbarMessage });
+    } catch (error) {
+      throw new Error(`${error}`);
```
CalebPena, line 83: Same thing as the email component. Can we add an error message to the snack bar?

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
```diff
@@ -0,0 +1,126 @@
...
+              helperText={
+                errors.phone !== undefined && (
+                  <ErrorMessageWrapper fontSize="1rem">{errors.phone.message}</ErrorMessageWrapper>
+                )
+              }
+              onChange={(...args) => {
+                field.onChange(...args);
+                if (isSubmitted) {
+                  trigger('phone');
+                }
```
CalebPena, line 112: Do we need this? I think that this happens automatically.

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
```diff
@@ -0,0 +1,126 @@
...
+              {...field}
+              label={<FormattedMessage id="signUp.createPhoneTextfield-label" defaultMessage="Cell Phone" />}
+              variant="outlined"
+              error={errors.phone !== undefined}
+              helperText={
+                errors.phone !== undefined && (
+                  <ErrorMessageWrapper fontSize="1rem">{errors.phone.message}</ErrorMessageWrapper>
+                )
+              }
+              onChange={(...args) => {
```
CalebPena, line 108: Can you use `handleNumbersOnly`, and `NUM_PAD_PROPS` link the sign up page, so the user can only enter numbers.

src/Components/SaveMyResultsModal/SaveMyResultsModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+type SaveMyResultsModalProps = {
+  close: () => void;
+};
+
+const SaveMyResultsModal = forwardRef(function SaveMyResultsModal({ close }: SaveMyResultsModalProps, ref) {
+  const intl = useIntl();
+  const [copied, setCopied] = useState(false);
+  const [snackbar, setSnackbar] = useState({
+    open: false,
+    message: '',
```
CalebPena, line 21: Can we make this a formatted message instead of a string? This has a bug where it won't be translated if the user does changes languages after it has been rendered. You can also make it an object with the default message and id.

src/Components/SaveMyResultsModal/EmailTextfield.tsx
ivonne-hernandez: Good idea! I've incorporated this in both the `EmailTextfield` and `PhoneTextfield` components.

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
ivonne-hernandez: It does. I removed it.

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
ivonne-hernandez: `Zod` applies transformations like `.transform()` through a `ZodEffects` object, and this breaks access to certain type-checking like .length() directly.

src/Components/SaveMyResultsModal/PhoneTextfield.tsx
ivonne-hernandez: @sydney-devine Can you please make this a separate ticket to refactor this number input as well as the one on the `SignUp` component:
Screenshot 2025-01-24 at 11 02 53 AM

CalebPena (approved):

PR #1471, opened 2025-01-28
Refactor/footer email

ivonne-hernandez:
- I refactored all of the email links to use `hello@myfriendben.org` instead.
- I refactored the `ResultsError` component to use the `email` value obtained from `config`.

- The React `EmailShareButton` wasn't working so I removed it.

Error page:
Screenshot 2025-01-28 at 2 17 15 PM

Buttons:
https://github.com/user-attachments/assets/f2fae1ab-c9e9-4651-ad28-66fa1c498315

### DIFF
--- a/src/Components/Results/ResultsError/ResultsError.tsx
+++ b/src/Components/Results/ResultsError/ResultsError.tsx
@@ -22,7 +22,7 @@ const ResultsError = () => {
             id="results-error.message"
             defaultMessage="We're sorry. We are having some trouble completing your request. Please make sure you have completed all of the questions on the screen and try again. If you are still unable to load your benefits results, please contact "
           />
-          {email}.
+          {email}
         </p>
         <Button
           className="error-button"
--- a/src/Components/Results/ResultsError/ResultsError.tsx
+++ b/src/Components/Results/ResultsError/ResultsError.tsx
@@ -3,9 +3,11 @@ import ErrorIcon from '../../../Assets/icons/error-icon.svg';
 import { useParams, useNavigate } from 'react-router-dom';
 import { FormattedMessage } from 'react-intl';
 import './ResultsError.css';
+import { useConfig } from '../../Config/configHook';
 
 const ResultsError = () => {
   const { uuid, whiteLabel } = useParams();
+  const { email } = useConfig('feedback_links');
   const navigate = useNavigate();
 
   return (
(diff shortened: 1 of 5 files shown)


src/Components/Results/ResultsError/ResultsError.tsx
```diff
@@ -18,8 +20,9 @@ const ResultsError = () => {
         <p className="error-message">
           <FormattedMessage
             id="results-error.message"
-            defaultMessage="We're sorry. We are having some trouble completing your request. Please make sure you have completed all of the questions on the screen and try again. If you are still unable to load your benefits results, please contact myfriendben@garycommunity.org."
+            defaultMessage="We're sorry. We are having some trouble completing your request. Please make sure you have completed all of the questions on the screen and try again. If you are still unable to load your benefits results, please contact "
           />
+          {email}.
```
CalebPena, line 25: Can we remove the `.`? Hard coding the `.` does not translate into every language. Also, Someone might try to email the email address plus the `.` at the end.

```suggestion
          {email}
```

src/Components/Share/Share.tsx
```diff
@@ -82,27 +87,27 @@ const Share = forwardRef(function Share() {
...
           </Icon>
         </FacebookShareButton>
-        <TwitterShareButton url={shareUrl} onClick={trackOutboundLinks('Share to Twitter')}>
+        <TwitterShareButton url={shareUrl} onClick={() => trackOutboundLinks('Share to Twitter')}>
           <Icon name="twitter">
             <XIcon sx={iconSize} />
           </Icon>
         </TwitterShareButton>
-        <EmailShareButton url={shareUrl} onClick={trackOutboundLinks('Share With Email')}>
+        <button onClick={handleEmailShare} className="email-hover">
```
CalebPena, line 100: We can make this an anchor tag instead. We do this in the footer.

CalebPena (changes requested):
The tracking needs to be fixed.

src/Components/Share/Share.tsx
ivonne-hernandez: 1. All of the icons are currently buttons
2. Making the email button an a tag means that I'll have to redo desktop and mobile styles
Screenshot 2025-01-28 at 2 55 16 PM

PR #1472, opened 2025-01-29
Electricity provider step

CalebPena:
- Add electricity provider step
- Add hook to initialize energy data
- Hard code the electricity provider by zip codes


src/Components/EnergyCalculator/Steps/ElectricityProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const energyDataAvailable = useEnergyFormData(formData);
+  const { uuid } = useParams();
+  const backNavigationFunction = useDefaultBackNavigationFunction('energyCalculatorElectricityProvider');
+  const nextStep = useGoToNextStep('energyCalculatorElectricityProvider');
+  const { formatMessage } = useIntl();
+  const { updateScreen } = useScreenApi();
+
+  const formSchema = z.object({
+    electricityProvider: z
+      .string({
```
ivonne-hernandez, line 32: What do you think about moving this inside of the min(1) argument and using a message object instead of an errorMap?

ivonne-hernandez (changes requested):
Please take a peek at the comment that I left regarding the error message rendering.

ivonne-hernandez (approved):

PR #1485, opened 2025-02-06
Gas provider step

CalebPena:
- Add gas provider step and gas provider on the confirmation page.


src/Components/EnergyCalculator/Steps/ElectricityProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+      .string({
+        errorMap: () => {
+          return {
+            message: formatMessage({
+              id: 'energyCalculator.electricityProvider.errorMessage',
+              defaultMessage: 'Please select an electricity provider',
+            }),
+          };
+        },
+      })
```
ivonne-hernandez, line 41: What do you think about moving this after the min(1) and inside of a normal message object?

ivonne-hernandez (changes requested):
Please take a peek at the comments that I left regarding the error messages.

ivonne-hernandez (approved):

PR #1497, opened 2025-02-12
Energy Calculator Results Page

CalebPena:
- Added feature flags.
- Add energy calculator results header.
- Add energy calculator results rebate cards.
- Add energy calculator results rebate page.
- Add required programs card to program page.

- Make some results components more reusable.


src/Assets/numInputHelpers.tsx
```diff
@@ -10,8 +10,11 @@ export const DECIMALS = /^\d*\.?\d*$/; // any numbers with unlimited decimal pla
 export const DOLLARS = /^\d*\.?\d{0,2}$/; // any numbers with a maximum of 2 decimal places
 
 // restrict what a user can enter into a text field
-export function handleNumbersOnly(onChange: ReactEventHandler<EventElements> = () => {}, regex: RegExp = INTEGERS) {
-  return (event: ChangeEvent<EventElements>) => {
+export function handleNumbersOnly<EventElement extends EventElements>(
+  onChange: (func: ChangeEvent<EventElement>) => void = () => {},
+  regex: RegExp = INTEGERS,
+) {
```
ivonne-hernandez, line 16: What do you think about making this type more readable? Does the onChange take a function that returns void which is an anonymous function that returns an empty object?

CalebPena: I extracted the type into its own type. `(func: ChangeEvent<EventElement>) => void` is the type, and `() => {}` is the default value if no function is passed in.

ivonne-hernandez (approved):

PR #1499, opened 2025-02-12
Feat/renter redirect pg

ivonne-hernandez:
- I added a `NoExpenses` component and css file to the `EnergyCalculator` directory. 
- I added a `no-expenses` route to `App`.

- I refactored the Energy Calculator's `Expenses` `formSubmitHandler` to navigate to the `no-expenses` route if the user did not select any expenses.

Screenshot 2025-02-12 at 2 47 20 PM

### DIFF
--- a/src/Assets/pageTitleTags.ts
+++ b/src/Assets/pageTitleTags.ts
@@ -10,6 +10,7 @@ export const QUESTION_TITLES: Record<QuestionName, string> = {
   acuteHHConditions: 'Near Term Help',
   referralSource: 'Referral',
   signUpInfo: 'Optional Sign Up',
+  energyCalculatorRedirectToMFB: 'Redirect to MFB',
 };
 
 export type OtherStepName = 'language' | 'disclaimer' | 'state' | 'confirmation' | 'results' | 'default';
--- a/src/Components/EnergyCalculator/NoExpensesPage/NoExpenses.tsx
+++ b/src/Components/EnergyCalculator/NoExpensesPage/NoExpenses.tsx
@@ -13,11 +13,14 @@ import './NoExpenses.css';
 const NoExpenses = () => {
   const { uuid } = useParams();
   const navigate = useNavigate();
+  const { formatMessage } = useIntl();
+  const errorIconAlt = formatMessage({ id: 'no-expenses-error-icon-alt', defaultMessage: 'error icon' });
   const { whiteLabel } = useContext(Context);
   const expensesStepId = useStepNumber('energyCalculatorExpenses');
 
-  const { formatMessage } = useIntl();
-  const errorIconAlt = formatMessage({ id: 'no-expenses-error-icon-alt', defaultMessage: 'error icon' });
+  useEffect(() => {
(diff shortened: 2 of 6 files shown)


src/Assets/pageTitleTags.ts
```diff
@@ -10,6 +10,7 @@ export const QUESTION_TITLES: Record<QuestionName, string> = {
   acuteHHConditions: 'Near Term Help',
   referralSource: 'Referral',
   signUpInfo: 'Optional Sign Up',
+  energyCalculatorRedirectToMFB: 'Redirect to MFB',
```
CalebPena, line 13: Add this to `OTHER_PAGE_TITLES`

src/Components/EnergyCalculator/NoExpensesPage/NoExpenses.tsx
```diff
@@ -0,0 +1,66 @@
+import { useContext, useEffect } from 'react';
+import { QUESTION_TITLES } from '../../../Assets/pageTitleTags';
+import QuestionHeader from '../../QuestionComponents/QuestionHeader';
+import { FormattedMessage, useIntl } from 'react-intl';
+import errorIcon from '../../../Assets/icons/error-icon.svg';
```
CalebPena, line 5: You can import this as a React component. Example: `/src/Components/Config/configHook.tsx`

CalebPena (approved):
Looks good after a couple of small fixes.

src/Components/EnergyCalculator/NoExpensesPage/NoExpenses.tsx
ivonne-hernandez: The issue with refactoring this into a React component is that it would require wrapping the `ErrorIcon` component in a div and assigning it a role="img" to provide an aria-label for accessibility. This adds unnecessary complexity, as the current implementation using the  tag is already semantic and fully accessible out of the box. The  tag natively supports the alt attribute, which is the standard and most straightforward way to provide accessibility for images. By keeping the  tag, we maintain simplicity, ensure consistent accessibility, and avoid introducing extra layers of markup and roles that don't provide any additional benefit.

src/Components/EnergyCalculator/NoExpensesPage/NoExpenses.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </div>
+      <div className="question-buttons">
+        <Button
+          variant="outlined"
+          onClick={backNavigationFunction}
+          startIcon={<NavigateBeforeIcon sx={{ mr: '-8px' }} className="rtl-mirror" />}
+        >
+          <FormattedMessage id="previousButton" defaultMessage="Back" />
+        </Button>
+        <Button variant="contained" href="/co/step-1?referrer=coenergycalculator" sx={{ textTransform: 'none' }}>
```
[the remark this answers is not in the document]
ivonne-hernandez: Thanks for the suggestion! I think sticking with `onClick` and `navigate` makes sense here since it allows us to handle the `uuid` check and maintain SPA behavior without a full page reload.

ivonne-hernandez:
The translations for this PR are in prod.

PR #1514, opened 2025-02-24
Feat/add 211 link

SonaliBedge:
--added 211nc link under Near-term-benefits tab

image (1)


src/Components/Referrer/Referrer.tsx
```diff
@@ -26,6 +27,16 @@ export const BrandedFooter = () => {
   }
   return <Footer />;
 };
+export const ResultsMessageForNeeds = () => {  
+  const { whiteLabel } = useParams();  
+  if (whiteLabel === 'nc')
```
CalebPena, line 32: Instead of hard coding the white label, can we add a feature flag? Then you can add the feature flag to the config.
```suggestion
  if (getReferrer('featureFlags').includes('nc_show_211_link'))
```

src/Components/Results/Link211Message.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default function Link211Message1() {
+  const { getReferrer } = useContext(Context);
+  const { needs } = useResultsContext();  
+  const Link211_Msg = getReferrer('link211Message');    
+  const defaultMessage = Link211_Msg.props.defaultMessage;  
+  return (  
+    <div
+      className="back-to-screen-message"
+      dangerouslySetInnerHTML={{ __html: defaultMessage }}
```
CalebPena, line 13: Lets avoid `dangerouslySetInnerHTML`, especially with text coming from the server (if someone was able to break into the admin and change the config, they would then be able to perform a cross site scripting attack). Instead, because we only will render this if a certain flag is present, lets rename this component to `NcLink211Message` and hard code the link in.

src/Components/Referrer/referrerHook.tsx
```diff
@@ -19,23 +23,26 @@ export type ReferrerData = {
   featureFlags: ReferrerOptions<string[]>;
   stepDirectory: ReferrerOptions<StepDirectory>;
   noResultMessage: ReferrerOptions<FormattedMessageType>;
+  link211Message?: ReferrerOptions<Link211MessageType> | null;
```
CalebPena, line 26: This is going to be handled by the feature flags instead, so you can remove it.

CalebPena (changes requested):
Can you use the feature flags instead?

src/Components/Results/NcLink211Message.tsx
```diff
@@ -5,12 +5,14 @@ import { useResultsContext } from './Results';
...
-  );  
+  const Link211_Msg = getReferrer('featureFlags');    
+    
+  return (
+    <div>
+      <p>{Link211_Msg.text}
+      <a href={Link211_Msg.url} target="_blank" rel="noopener noreferrer">
+        Click here
+      </a></p>
+    </div>
```
CalebPena, line 16: Because this is under a NC feature flag, please just hard code the text into the FE. No need to add it to the config because this will only show up for NC.

CalebPena (changes requested):
Can you hard code the text because this message only shows up for NC.

src/Components/Results/NcLink211Message.tsx
```diff
@@ -1,17 +1,13 @@
...
+  const { needs } = useResultsContext();   
     
   return (
     <div>
-      <p>{Link211_Msg.text}
-      <a href={Link211_Msg.url} target="_blank" rel="noopener noreferrer">
-        Click here
+      <p>For more information or to learn more, please visit the following link  :  
+      <a href="https://nc211.org/" target="_blank" rel="noopener noreferrer">
+         Click here
```
CalebPena, line 10: Can you please turn these into a `<FormattedMessage />` so that they can be translated.

CalebPena (changes requested):
One more change.

src/Components/Results/NcLink211Message.tsx
SonaliBedge: Please let me know if any changes are needed, even if the message should be completely different.

CalebPena (approved):
Looks good.

PR #1517, opened 2025-02-25
Updating form data from api responses

kbrey:
### What (if any) features are you implementing?
- Updates form data from api responses, rather than updating form data manually (fixes issue https://github.com/Gary-Community-Ventures/benefits-api/issues/717)

We already had code to support the user refreshing the page, in `FetchScreen`, which fetched the screen from the api and populated the form data. This PR refactors that code, so that it can also be used to sync the frontend state when we create or update the screen.

Now, whenever the api is called to update the screen, the formData is updated. This also means when each step is submitted, we don't have to both update the formData and call the update api, so I've also removed those `setFormData` calls from the steps.

### Will this impact UX?
- When the user submits a page, do not wait for the update screen response before moving to the next page, so nothing should appear slower for the user.

### Notes
- I found an issue where if you refresh the page, then edit a household member, the white label goes to _default. This is because HHMSummaryCards relies on formData.whiteLabel which is not populated when the screen is fetched (instead it is set when the user selects their language). I fixed this by removing the "whiteLabel" field from being set or referenced because it didn't appear to be used anywhere else, and am instead using the context white label.
- I removed the DropDownMenu components because they set the form data and didn't appear to be used.


SonaliBedge:
> What (if any) features are you implementing?
> 
> * Updates form data from api responses, rather than updating form data manually
> 
> We already had code to support the user refreshing the page, in `FetchScreen`, that fetches the screen from the api and populates the form data. This PR refactors that code, so that it can also be used when we create or update the screen to sync the frontend state.
> 
> Now, whenever the api is called to update the screen, the formdata is updated. This also means when each step is submitted, we don't have to both update the formdata and call the update api, so I've also removed those `setFormData` calls.
> 
> Remaining tasks
> 
> * Address any TODOs
> * Make sure that formData is only updated from api responses if possible

I will continue working on the remaining tasks mentioned above.

src/Components/PreviousButton/PreviousButton.tsx
```diff
@@ -21,7 +21,7 @@ const PreviousButton = ({ navFunction }: Props) => {
 
   const defaultNavigate = () => {
     if (id && +id === householdStep + 1) {
-      navigate(`/${whiteLabel}/${uuid}/step-${householdStep}/${formData.householdData.length}`);
+      navigate(`/${whiteLabel}/${uuid}/step-${householdStep}/${formData.householdData.length || 1}`);
```
CalebPena, line 24: Can we use `??`. I prefer it because it does not change behavior for `0` and `''`.
```suggestion
      navigate(`/${whiteLabel}/${uuid}/step-${householdStep}/${formData.householdData.length ?? 1}`);
```

src/Components/QuestionComponents/questionHooks.ts
```diff
@@ -69,7 +69,7 @@ export function useDefaultBackNavigationFunction(questionName: QuestionName) {
 
   const prevUrl = useMemo(() => {
     if (prevStepName && MEMBER_QUESTIONS.includes(prevStepName)) {
-      return `/${whiteLabel}/${uuid}/step-${currentStepId - 1}/${formData.householdData.length}`;
+      return `/${whiteLabel}/${uuid}/step-${currentStepId - 1}/${formData.householdData.length || 1}`;
```
CalebPena, line 72:
```suggestion
      return `/${whiteLabel}/${uuid}/step-${currentStepId - 1}/${formData.householdData.length ?? 1}`;
```

CalebPena (approved):
LGTM. Left 2 nits.

src/Components/PreviousButton/PreviousButton.tsx
kbrey: I used `||` intentionally to handle the 0 case. In the off chance the user goes back a page before the form data updates, the householdData length was 0, which caused a weird routing bug, so wanted it to default to 1 in that case.

PR #1530, opened 2025-02-27
Fix copy on Energy Calculator confirmation page

CalebPena:
- Rename the expense question title on the confirmation page
- Add "None" if the household member has no conditions


CalebPena:
Done

PR #1546, opened 2025-03-04
Add state to zip code error message

CalebPena:
- Add state specific zip code error message.
- Add description that links to how to change your state.

image

- Fix reloading on the state page.

**I think that we should release this when NC goes live, so we don't link to a page with one state option.**

**I can change the select you state link easily**

### DIFF
--- a/src/Components/Steps/Zipcode.tsx
+++ b/src/Components/Steps/Zipcode.tsx
@@ -17,14 +17,17 @@ import PrevAndContinueButtons from '../PrevAndContinueButtons/PrevAndContinueBut
 import { useDefaultBackNavigationFunction, useGoToNextStep } from '../QuestionComponents/questionHooks';
 import { handleNumbersOnly, NUM_PAD_PROPS } from '../../Assets/numInputHelpers';
 import useScreenApi from '../../Assets/updateScreen';
+import QuestionDescription from '../QuestionComponents/QuestionDescription';
 
 export const Zipcode = () => {
-  const { formData, setFormData } = useContext(Context);
+  const { formData, setFormData, getReferrer } = useContext(Context);
   const { uuid } = useParams();
   const backNavigationFunction = useDefaultBackNavigationFunction('zipcode');
   const { updateScreen } = useScreenApi();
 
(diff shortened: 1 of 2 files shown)


src/Components/Steps/Zipcode.tsx
```diff
@@ -17,14 +17,17 @@ import PrevAndContinueButtons from '../PrevAndContinueButtons/PrevAndContinueBut
...
+import QuestionDescription from '../QuestionComponents/QuestionDescription';
 
 export const Zipcode = () => {
-  const { formData, setFormData } = useContext(Context);
+  const { formData, setFormData, getReferrer } = useContext(Context);
   const { uuid } = useParams();
   const backNavigationFunction = useDefaultBackNavigationFunction('zipcode');
   const { updateScreen } = useScreenApi();
 
+  const noChangeStateMessage = getReferrer('featureFlags').includes('no_zipcode_change_state');
```
ivonne-hernandez, line 28: Can you please help me understand what this does?

src/Components/Steps/Zipcode.tsx
```diff
@@ -128,6 +139,20 @@ export const Zipcode = () => {
       <QuestionQuestion>
         <FormattedMessage id="questions.zipcode" defaultMessage="What is your zip code?" />
       </QuestionQuestion>
+      {!noChangeStateMessage && (
```
kbrey, line 142: You mentioned in the PR description to wait to merge this, but since this is in a feature flag couldn't we go ahead and merge but have it off?

kbrey (approved):
LGTM!

ivonne-hernandez:
Feature flags are a temporary tool designed to enable faster, safer deployments, facilitate testing, and support gradual rollouts, but they are not meant to be permanent. Over time, they can introduce technical debt, clutter the codebase, and increase complexity if not managed properly. Once a feature is fully rolled out and stable, the flag and associated conditional logic should be removed, and the codebase should be refactored into a more maintainable structure. This ensures long-term clarity and reduces the risk of unnecessary complexity. Let’s plan to clean up this flag once the feature is confirmed stable.

PR #1547, opened 2025-03-04
Redirect white label route to step one.

CalebPena:
- Redirect `/:whiteLabel` to `/:whiteLabel/step-1`


src/Types/WhiteLabel.ts
```diff
@@ -0,0 +1,2 @@
+export const ALL_VALID_WHITE_LABELS = ['co', 'nc', 'co_energy_calculator'] as const;
+export type WhiteLabel = (typeof ALL_VALID_WHITE_LABELS)[number];
```
ivonne-hernandez, line 2: @CalebPena What do you think about using `export type WhiteLabel = 'co' | 'nc' | 'co_energy_calculator'` instead for readability? If other values need to be added to the `ALL_VALID_WHITE_LABELS` type it's super quick to add those to the `WhiteLabel` type as well.

ivonne-hernandez (changes requested):
@CalebPena What do you think about using export type WhiteLabel = 'co' | 'nc' | 'co_energy_calculator' instead for readability? If other values need to be added to the ALL_VALID_WHITE_LABELS type it's super quick to add those to the WhiteLabel type as well.

ivonne-hernandez (approved):

PR #1564, opened 2025-03-12
Add message about urgent needs benefits on NC 211 co-brand

IrynaKolh:

- Add message about urgent needs benefits on NC 211 co-brand

image


src/Components/Results/NcLink211Message.tsx
```diff
@@ -1,15 +1,15 @@
 import { FormattedMessage } from 'react-intl';
 
-export default function Link211Message1() {
+export default function NcLink211Message() {
```
CalebPena, line 3: I think that you should also change the name where this is imported.

PR #1607, opened 2025-04-09
MA: Massachusetts

CalebPena:
- Add veteran services urgent need category.


src/Components/Steps/Disclaimer/Disclaimer.tsx
```diff
@@ -108,7 +108,6 @@ const Disclaimer = () => {
                 className="link-color"
                 target="_blank"
                 href={publicChargeOption.link}
-                target="_blank"
```
kbrey, line 111: Wasn't this change added recently? Just want to make sure removing this is intentional.

CalebPena: It is in there twice. See 2 lines above

kbrey: Got it.

kbrey (approved):

PR #1619, opened 2025-04-15
Central: Members caps

CalebPena:
- Extend cap functionality to individual members.


myang1010:
Looks like this was already merged, sorry for the delay! Took me a bit to wrap my head around the code and understand how it works. Looks good!

Issue #1606, opened 2025-04-08
CENTRAL: Race condition when updating form data


CalebPena:
Now that the form data gets updated after each step, there is a race condition where it is possible to override the previous step if the next steps request happens before the previous request is done. Only happens if you are going through very fast.

https://github.com/user-attachments/assets/2d3798a9-4465-41a0-a565-eda711476ee7

PR #1625, opened 2025-04-18
Bug: formdata race condition

kbrey:
- Fixes (https://github.com/Gary-Community-Ventures/benefits-calculator/issues/1606)

- I updated the submission handler in each step to wait for the updateScreen api call before navigating to the next step.
- I created a helper hook, `useStepForm` that wraps useForm. This hook automatically calls `nextStep` and sets the stepLoading context value so that we display loading spinner within the continue button. In the future we could consider a better UI.

- I considered other options that didn't require the UX to wait for the api call. However, they introduced more complexity than added value in my opinion.

- I tested the NC and CO Energy white labels as best as I could but would like feedback on if I missed any cases, as this change will have a wide impact.

-


src/Components/Steps/SignUp/SignUp.tsx
```diff
@@ -221,9 +223,9 @@ function SignUp() {
       await updateUser(newFormData);
       newFormData.signUpInfo.hasUser = true;
       setFormData(newFormData);
-      nextStep();
-    } catch {
+    } catch (e) {
       setHasServerError(true);
+      throw e;
```
kbrey, line 228: I did test locally that if an error is thrown, the submit successful state is not triggered.

myang1010 (commented):
Thanks Katie for taking this on! I'm following along and getting a better sense of the direction. It seems like the goal is to ensure each step completes then move to the next.

Just wanted to share a small observation from testing locally: I noticed a minor thing that the "Continue" text on the button briefly disappears right after clicking, looks like it's from the `stepLoading` state. I’ve attached a video in case it helps clarify what I mean.

Also, just curious, since the script auto-population is mainly for developer convenience, would it be possible to slow it down slightly to avoid these changes? Totally get it if the current approach is intended to improve the screener’s robustness. Just wanted to raise the thought. Thanks again!

https://github.com/user-attachments/assets/fd843127-017a-42f9-ad66-d18f7ff8b948

src/Components/Steps/stepForm.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { QuestionName } from '../../Types/Questions';
+import { useGoToNextStep } from '../QuestionComponents/questionHooks';
+
+// This hook is used to create a form for screener steps.
+// By default, it will navigate to the next step when the form is submitted successfully.
+// The onSubmitSuccessfulOverride callback should be used when custom logic is needed.
+//
+// If you must navigate within the submit handler, use the setStepLoading function
+// to set the loading state to false before navigating. Otherwise, the isSubmitSuccessful
+// state will not be updated correctly and the loading state will not be set to false.
```
CalebPena, line 13: Can you make this a block comment so that it shows up when you hover the function?

src/Components/ContinueButton/FormContinueButton.tsx
```diff
@@ -1,11 +1,17 @@
...
-import { PropsWithChildren } from 'react';
+import { Button, CircularProgress } from '@mui/material';
+import { PropsWithChildren, useContext } from 'react';
 import { FormattedMessage } from 'react-intl';
+import { Context } from '../Wrapper/Wrapper';
 
 const FormContinueButton = ({ children }: PropsWithChildren) => {
+  const { stepLoading } = useContext(Context);
   return (
     <Button variant="contained" type="submit">
```
CalebPena, line 9: Can we make the box stay the same size when you submit?

CalebPena (approved):
Looks good. Just a couple of nits.

src/Components/EnergyCalculator/Steps/Disclaimer.tsx
```diff
@@ -65,32 +65,38 @@ const Disclaimer = () => {
...
 
+  type FormSchema = z.infer<typeof formSchema>;
+
   const {
     control,
     formState: { errors },
     getValues,
     handleSubmit,
-  } = useForm<z.infer<typeof formSchema>>({
+  } = useForm<FormSchema>({
```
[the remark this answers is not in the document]
kbrey: I had that initially but removed it because the Disclaimer component doesn't use the `useGoToNextStep` hook. So it requires some special overrides. I do like it being consistent though so I am definitely open to adding it back.

src/Components/ContinueButton/FormContinueButton.tsx
kbrey: Great idea, done.

src/Components/Steps/stepForm.tsx
kbrey: Done.

kbrey:
> Thanks Katie for taking this on! I'm following along and getting a better sense of the direction. It seems like the goal is to ensure each step completes then move to the next.
> 
> Just wanted to share a small observation from testing locally: I noticed a minor thing that the "Continue" text on the button briefly disappears right after clicking, looks like it's from the `stepLoading` state. I’ve attached a video in case it helps clarify what I mean.
> 
> Also, just curious, since the script auto-population is mainly for developer convenience, would it be possible to slow it down slightly to avoid these changes? Totally get it if the current approach is intended to improve the screener’s robustness. Just wanted to raise the thought. Thanks again!
> 
>  continue_btn_disappear_briefly.mov

Thanks @myang1010! 
- I think the button "disappearing" that you are seeing is because the width was changing when the loading spinner displayed. I just made it so the width stays the same so hopefully that's fixed now - can you please check?
- You are right that this issue only happens with the script now - but I agree with your other point that this will make the screener more robust and avoid weird bugs that could theoretically happen.

src/Components/ContinueButton/FormContinueButton.tsx
myang1010: Just wondering, could we use `disabled={stepLoading}` here? If so, we might not need the following changes.

src/Components/EnergyCalculator/Steps/Disclaimer.tsx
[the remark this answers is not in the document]
CalebPena: Didn't you add the ability to provide an override function in place of the `useGoToNextStep` hook. Couldn't you use that?

src/Components/Steps/Disclaimer/Disclaimer.tsx
```diff
@@ -68,32 +68,40 @@ const Disclaimer = () => {
...
 
-  const formSubmitHandler: SubmitHandler<z.infer<typeof formSchema>> = async (termsOfServiceAndAgeData) => {
+  const formSubmitHandler: SubmitHandler<FormSchema> = async (termsOfServiceAndAgeData) => {
     const updatedFormData = { ...formData, ...termsOfServiceAndAgeData };
+    setStepLoading(true);
 
     if (uuid) {
       await updateScreen(updatedFormData);
-      navigate(`/${whiteLabel}/${uuid}/step-${STARTING_QUESTION_NUMBER}`);
+      startScreen(uuid);
```
CalebPena, line 92: Do you need to set screen loading here?

src/Components/EnergyCalculator/Steps/Disclaimer.tsx
```diff
@@ -65,32 +65,40 @@ const Disclaimer = () => {
...
 
-  const formSubmitHandler: SubmitHandler<z.infer<typeof formSchema>> = async (termsOfServiceAndAgeData) => {
+  const formSubmitHandler: SubmitHandler<FormSchema> = async (termsOfServiceAndAgeData) => {
     const updatedFormData = { ...formData, ...termsOfServiceAndAgeData };
+    setStepLoading(true);
 
     if (uuid) {
       await updateScreen(updatedFormData);
-      navigate(`/co_energy_calculator/${uuid}/step-${STARTING_QUESTION_NUMBER}`);
+      startScreen(uuid);
```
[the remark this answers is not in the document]
kbrey: I believe I didn't change any calls to setScreenLoading, the behavior there should be the same as before where `setScreenLoading(false)` is called only when the screen needs to be created. Is that what you meant?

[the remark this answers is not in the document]
CalebPena: Oops. I was confusing this with `setStepLoading`

CalebPena (approved):
LGTM

PR #1627, opened 2025-04-18
Central: Correct zip code error message

CalebPena:
- Fix zip code message.

### DIFF
--- a/src/Components/Steps/Zipcode.tsx
+++ b/src/Components/Steps/Zipcode.tsx
@@ -46,7 +46,7 @@ export const Zipcode = () => {
       helperText(
         formatMessage({
           id: 'validation-helperText.zipcode',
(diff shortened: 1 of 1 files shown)


myang1010 (approved):
Thanks!

src/Components/Steps/Zipcode.tsx
```diff
@@ -46,8 +46,8 @@ export const Zipcode = () => {
       helperText(
         formatMessage({
           id: 'validation-helperText.zipcode',
-          defaultMessage: `Please enter a valid zip code for ${state.name}`,
-        }),
+          defaultMessage: `Please enter a valid zip code for `,
```
myang1010, line 49: For consistency, maybe use `'` instead of `` ` ``?

CalebPena: Fixed

PR #1643, opened 2025-04-28
Central: Add FPL utility function to dev console

CalebPena:
- Add a utility function to the dev console to easily get the monthly value for a FPL percent.

image


myang1010 (approved):
Thanks for adding this tool! Super helpful for getting a quick idea of monthly FPL by household size.

PR #1644, opened 2025-04-28
Feature/End-to-end testing with Playwright

msrezaie:
- Added Playwright package for performing end-to-end testing.
- This PR also includes two sample tests: 1) `basic-smoke.spec.ts` a basic smoke test (which only tests that the app loads fine). 2) `sample-e2e.spec.ts` a sample end to end screen test (this test runs from the initial load of the app until the results page)

- Removed the `cypress` package.
- Refactored some of the `select` elements' `id` attributes to have better names. This helps in better targeting of elements with test files.

- I've updated the readme file with instructions on how to set up, run, and create new tests.


tests/sample-e2e.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Sample screen test: end-to-end
+ */
+test.describe('MyFriendBen Sample Screen Test', () => {
+  test('start to finish screen test', async ({ page }) => {
+    await page.goto('/');
+    await page.getByRole('button', { name: 'Get Started' }).click();
+
+    await expect(page).toHaveURL('/select-state');
+    await page.getByLabel('', { exact: true }).click();
```
CalebPena, line 12: Is the label an empty string?

tests/sample-e2e.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    await page.getByRole('option', { name: 'Test / Prospective Partner' }).click();
+    await page.getByRole('button', { name: 'Continue' }).click();
+    
+    await expect(page).toHaveURL(/\/nc\/.*\/step-11/);
+    await page.getByRole('button', { name: 'Continue' }).click();
+    
+    await expect(page).toHaveURL(/\/nc\/.*\/confirm-information/);
+    await page.getByRole('button', { name: 'Continue' }).click();
+    await page.getByRole('button', { name: 'save my results' }).click();
```
CalebPena, line 91: Do you want to make sure that it gets to the results page?

tests/sample-e2e.spec.ts
msrezaie: Thanks for pointing this out. I've now made the targets more specific.

CalebPena (approved):
LGTM

PR #1647, opened 2025-04-29
Feat/show-Medicaid-for-all-eligible-household-members

myang1010:
- This change will make sure Medicaid shows up correctly for the rest of the household without it.

- Skip adding the household value if a member already has the program
- Update the `doesNotHave` logic. Check member level if the program is Medicaid.
- Update `MemberEligibility` and include `already_has`

- Any thoughts on the change please let me know


src/Components/Results/FormattedValue.tsx
```diff
@@ -9,6 +9,9 @@ export function programValue(program: Program) {
   let total = program.household_value;
 
   for (const member of program.members) {
+    if (member.already_has)
+      continue
```
CalebPena, line 13: Nit
```suggestion
    if (member.already_has) {
      continue
    }
```

src/Components/Results/filterPrograms.ts
```diff
@@ -84,7 +84,11 @@ export default function filterProgramsGenerator(
     const meetsLegalStatus = program.legal_status_required.some((status) => filtersCheckedStrArr.includes(status));
     const isEligible = program.eligible;
     const hasValue = programValue(program) > 0;
-    const doesNotHave = !program.already_has;
+    let doesNotHave = !program.already_has;
+
+    if (program.name_abbreviated.includes("medicaid")){
+      doesNotHave = program.members.some(member => member.already_has === false);
+    }
```
CalebPena, line 91: You don't need this because it should be handled in the `programValue`
```suggestion
```

CalebPena (changes requested):

src/Components/Results/filterPrograms.ts
myang1010: I think this line `has_insurance = self.has_insurance_types((name_abbreviated,), strict=False)`  in the `Screen` model's `has_benefit()` is causing it to return `true`. Without above added lines, Medicaid doesn't show on the result page. If I remove the insurance check, will it affect how other programs are display?

CalebPena: We should move the insurance check to the member `has_benefit`. It is really only there for Medicaid, but I would do a quick search through the code base, to make sure we are not using `has_benefit` for any of the other insurance options.

CalebPena (approved):

PR #1654, opened 2025-04-30
Fix Typescript Errors: Part 2 - Assets, Types, tsconfig

IrynaKolh:
- Fix Typescript Errors: Part 2 - Assets, Types, tsconfig

Files
2 src/Assets/analytics.ts:4
1 src/Assets/updateFormData.tsx:23
1 src/Assets/updateScreen.ts:32
1 src/Types/Results.ts:2
1 tsconfig.json:18


tsconfig.json
```diff
@@ -14,8 +14,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx",
-    "allowImportingTsExtensions": true
```
kbrey, line 18: @IrynaKolh this does fix the error when running tsc, but I am now seeing more errors when I run the app related to file extensions. To limit the scope of this ticket I would suggest to add back this `"allowImportingTsExtensions": true` and we can make another ticket to clean that up later if needed.

IrynaKolh: ok

src/Types/FormData.ts
```diff
@@ -84,7 +84,7 @@ export type FormData = {
   startTime: string;
   hasExpenses: string;
   expenses: Expense[];
-  householdSize: number;
+  householdSize: number | string;
```
kbrey, line 87: Using the type `number | string` does fix the errors listed in the ticket. But I'm wondering if we should keep the type to "number" because it is set as a number in [HouseholdSize.tsx](https://github.com/Gary-Community-Ventures/benefits-calculator/blob/434e3a94538c0cdb648a7ab344214cd36e60cfd9/src/Components/Steps/HouseholdSize.tsx#L67). In that case I think you'll need to fix updateFormData to not set householdSize to a string. You will also need to check that the update screen api request will still work if householdSize is sent as a number.

IrynaKolh: I fixed again)
I checked - no errors in UI

src/Assets/updateFormData.tsx
```diff
@@ -20,7 +20,7 @@ export function useUpdateFormData() {
       startTime: response.start_date ?? formData.startTime,
       hasExpenses: response.expenses.length ? 'true' : 'false',
       expenses: [],
-      householdSize: String(response.household_size ?? ''),
+      householdSize: Number(response.household_size ?? ''),
```
CalebPena, line 23:
```suggestion
      householdSize: Number(response.household_size ?? 0),
```

CalebPena, line 23:
```suggestion
      householdSize: Number(response.household_size ?? 0),
```

CalebPena (approved):
Can you make one other small change?

PR #1666, opened 2025-05-05
Central: Add a default icon for current benefits categories

CalebPena:
- Add a default icon fall back for the current benefits category.

- Remove unnecessary logic.

**We should think about adding an urgent need category model like we do for the programs.**


myang1010 (approved):
Looks good to me.

kbrey:
> What (if any) features are you implementing?
> 
> * Add a default icon fall back for the current benefits category.
> 
> What (if anything) did you refactor?
> 
> * Remove unnecessary logic.
> 
> **We should think about adding an urgent need category model like we do for the programs.**

I like the suggestion to add a category model for urgent needs.

PR #1688, opened 2025-05-13
Fix immigration filter scrolling

SonaliBedge:
-Fixes the scroll event handling to prevent it from closing when scrolling inside filter in smaller windows
-Update the CSS to properly handle scrolling and heights
-Add proper scroll containment
-Fix popover positioning

### DIFF
--- a/src/Components/Results/Programs/Filter.tsx
+++ b/src/Components/Results/Programs/Filter.tsx
@@ -151,18 +151,16 @@ export const Filter = () => {
     setFilterHeight(0);
   };
 
-  const handleScroll = () => {
-    if (citizenshipFilterIsOpen) {
+  const handleScroll = (event: Event) => {
+    if (!citizenshipFilterIsOpen) return;    
+    
+    const target = event.target as Node;
+    const filterContainer = filterRef.current;
+    
+    if (filterContainer && !filterContainer.contains(target) && target !== filterContainer) {
       handleFilterClose();
     }
   };
-
-  useEffect(() => {
-    window.addEventListener('scroll', handleScroll);
-    return () => {
-      window.removeEventListener('scroll', handleScroll);
-    };
-  }, [citizenshipFilterIsOpen]);
 
   const citizenshipFiltersModalALProps = {
     id: 'filter.citFilterModalAL',
@@ -191,6 +189,12 @@ export const Filter = () => {
           }}
           transformOrigin={{ vertical: 2, horizontal: 0 }}
           aria-label={intl.formatMessage(citizenshipFiltersModalALProps)}
+          keepMounted
+          disableScrollLock={false}
+          style={{ position: 'fixed' }}
+          PaperProps={{
+            style: { overscrollBehavior: 'contain' }
+          }}
         >
           <div className="filters-close-button">
             <IconButton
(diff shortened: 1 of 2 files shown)


src/Components/Results/Programs/Filter.tsx
```diff
@@ -151,19 +151,17 @@ export const Filter = () => {
     setFilterHeight(0);
   };
 
-  const handleScroll = () => {
-    if (citizenshipFilterIsOpen) {
+  const handleScroll = (event: Event) => {
```
msrezaie, line 154: is this function being used in anywhere?

msrezaie (approved):
looks good, except for that `handleScroll` where its probably not being used in anywhere. So you might want to remove that.

src/Components/Results/Programs/Filter.tsx
```diff
@@ -191,6 +178,12 @@ export const Filter = () => {
           }}
           transformOrigin={{ vertical: 2, horizontal: 0 }}
           aria-label={intl.formatMessage(citizenshipFiltersModalALProps)}
+          keepMounted
+          disableScrollLock={false}
+          style={{ position: 'fixed' }}
+          PaperProps={{
+            style: { overscrollBehavior: 'contain' }
+          }}
```
CalebPena, line 186: Do we need these?

src/Components/Results/Programs/Filter.tsx
```diff
@@ -151,19 +151,6 @@ export const Filter = () => {
...
-      handleFilterClose();
-    }
-  };
-
-  useEffect(() => {
-    window.addEventListener('scroll', handleScroll);
-    return () => {
-      window.removeEventListener('scroll', handleScroll);
-    };
-  }, [citizenshipFilterIsOpen]);
```
CalebPena, line 165: I thought that this was the only piece that we need to change. What do the other pieces do?

CalebPena (changes requested):

src/Components/Results/Programs/Filter.tsx
SonaliBedge: We don't need this code — I missed removing it. The first time I tried changing the CSS in the filter.css file, it didn’t work, so I added the style in filter.tsx just to test if it would work, based on Claude’s suggestion. Later, I forgot to remove it. I’ll go ahead and remove it now. I’ve tested the screener after removing the code. Thank you, @CalebPena .

src/Components/Results/Programs/Filter.tsx
SonaliBedge: The handleScroll() function was present before but not used anywhere in the Benefits Calculator, so I removed it. To fix the immigration filter scrolling issue, I made changes and added code only in the CSS file.

src/Components/Results/Programs/Filter.tsx
CalebPena: Do you need the CSS you added? I thought that you just needed to remove the `handleScroll` stuff.

CalebPena:
@SonaliBedge Oops, I accidentally merged this. The only other question I have is about the CSS.

SonaliBedge:
@CalebPena , I tried modifying the handleScroll() function to fix the scroll issue with the immigration filter, but it didn’t work. However, when I changed the filter.css file, it worked.

CalebPena:
@SonaliBedge I think this is good.

PR #1700, opened 2025-05-15
Energy: Update renter no expense path

CalebPena:
- Remove path for renter with no expenses.


kbrey (approved):
Looks good.

PR #1702, opened 2025-05-15
Energy: Landing page logo link

CalebPena:
- Add feature flag to make the header logo link go to the landing page.

### DIFF
--- a/src/Components/Header/Header.tsx
+++ b/src/Components/Header/Header.tsx
@@ -24,6 +24,10 @@ const Header = () => {
       return `/step-1${queryString}`;
     }
 
+    if (getReferrer('featureFlags').includes('logo_landing_page_link')) {
+      return `/${whiteLabel}/landing-page${queryString}`;
+    }
+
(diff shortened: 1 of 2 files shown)


src/Components/Header/Header.tsx
```diff
@@ -24,6 +24,10 @@ const Header = () => {
       return `/step-1${queryString}`;
     }
 
+    if (getReferrer('featureFlags').includes('logo_landing_page_link')) {
+      return `/${whiteLabel}/landing-page${queryString}`;
```
kbrey, line 28: I'm curious about the decision to put the renter/owner selection on a landing page vs within the screener itself?

kbrey (approved):
Looks good.

src/Components/Header/Header.tsx
CalebPena: They have different questions, and the easiest way we could think of to have different paths was to make that the first thing that we do.

PR #1705, opened 2025-05-16
Feat/E2E test for current-benefits pages

msrezaie:
- Added a new test `current-benefits.spec` to test for `[PLACEHOLDER]` and untranslated labels in all of the white-labels current-benefits pages.

- Refactored and added steps to better clarify the test process in the NC e2e test `nc-end-to-end.spec` (originally called `sample-e2e.spec`).


tests/current-benefits.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+          return textNodes;
+        });
+
+        const untranslatedLabelPatterns = [
+          /program\.[a-z0-9_-]+/,
+          /program_category\.[a-z0-9_-]+/
+        ];
+
+        for (const text of textContent as string[]) {
```
CalebPena, line 73: Do you need the `as string[]` part. The function returns a `string[]`, so I feel like the type should be inferred.
```suggestion
        for (const text of textContent) {
```

CalebPena (approved):
Looks good besides a couple of changes.

CalebPena (approved):
LGTM

PR #1709, opened 2025-05-19
Energy: Renter path bug on landing page

CalebPena:
- Add ability to force remove param.
- Don't keep the path param when going to the landing page.
- Don't have a path param when clicking the homeowner button.


jmejia (approved):
New to the codebase so please, someone else chime in if you have thoughts, but this looks like it should be backwards compatible and type-safe based on my limited understanding.

kbrey (approved):
Cool looks good.

PR #1742, opened 2025-05-27
Refactor Playwright Helpers #1676 

mexi-cano:
- Refactor Playwright tests to use shared, type-safe helper functions.
- Create modular helper structure with selectors, navigation & form helpers
- Implement reusable flow abstractions for common test patterns
- Add white-label specific flows for NC implementation

- You can run all the tests with `npx playwright test` and the NC tests specifically with `npx playwright test tests/nc-end-to-end.spec.ts`
- There are two tests that fail. These are expected since the current benefits page has `PLACHOLDER` text and an untranslated label.


tests/helpers/flows/base.ts
kbrey, line 1: I think these base helpers are not used?

tests/helpers/flows/types.ts
kbrey, line 1: I think these helpers are not used?

tests/helpers/flows/nc.ts
```diff
@@ -0,0 +1,198 @@
...
+ * @param page - Playwright page instance
+ * @param data - Application data to use for the test
+ * @returns Promise with flow result
+ */
+export async function runNcEndToEndTest(
+  page: Page,
+  data: ApplicationData
+): Promise<FlowResult> {
+  // Add retry mechanism for more resilience
+  const maxRetries = 2;
```
kbrey, line 144: I guess I'm thinking rather than add a retry mechanism I would opt for adding more specific waits. If we do decide to go with a retry mechanism we could move this somewhere common.

tests/helpers/flows/nc.ts
```diff
@@ -0,0 +1,198 @@
...
+      // Setup session (landing page, state selection, disclaimer)
+      const setupResult = await setupNcSession(page);
+      if (!setupResult.success) {
+        lastError = setupResult.error || new Error('Unknown error in setup');
+        attempt++;
+        continue;
+      }
+      
+      // Complete full application with additional waiting for stability
+      await page.waitForLoadState('networkidle');
```
kbrey, line 169: In the docs I see useing networkidle is discouraged, any thoughts if it is worth it here?

tests/helpers/utils/constants.ts
```diff
@@ -0,0 +1,53 @@
...
+  REFERRAL_SOURCE: /step-10/,
+  ADDITIONAL_INFO: /step-11/,
+  CONFIRM_INFORMATION: /confirm-information/,
+  RESULTS: /results\/benefits/,
+};
+
+/**
+ * White label identifiers
+ */
+export const WHITE_LABELS = {
```
kbrey, line 31: Our white label list should be "nc", "co", "co_energy_calculator", and "ma", the others here are referrers I believe.

kbrey (changes requested):
@mexi-cano there's some great set up here! I have some suggestions for you, mostly around cleanup.

tests/nc-end-to-end.spec.ts
```diff
@@ -1,10 +1,15 @@
 import { test, expect } from '@playwright/test';
+import { runNcEndToEndTest, verifyCurrentUrl, saveResults } from './helpers';
+import { testUsers } from './helpers/utils/test-data';
```
kbrey, line 3: @mexi-cano It looks like these imports are not used. testUsers is particular looks like it could be useful here to fill in the form values, I would suggest we either use it or delete it.

kbrey (approved):
@mexi-cano great work! Thanks for the iterations, looks good to me.

Issue #1004, opened 2024-04-02
Update income limits for Health First Colorado / Medicaid to appy 2024 FPL


kbottagaro:
Update FPL applicable to Health First Colorado / Medicaid in Policy Engine to use 2024 FPL.

Background: HCPF updated Medicaid income limits as of April 1 to use 2024 FPL. The updated Medicaid income guidelines are at https://hcpf.colorado.gov/sites/hcpf/files/April%202024%20Medicaid%20Income%20Chart.pdf. The HCPF chart doesn't reflect the 5% income disregard, but that should already be taken into account in our calculations, so rules should work in Policy Engine so long as FPL is updated to reflect 2024.

PR #1765, opened 2025-06-09
Refactor/For the new UrgentNeedType model

msrezaie:
- Refactored `getAllNearTermPrograms` component to use the new endpoint for `UrgentNeedType`.
- Refactored the Near-Term Benefits (urgent need programs) section in the results page to use `iconCategoryMap`. This makes it so that all icon assets are available to urgent need types.

Related benefits-api PR: [#1004](https://github.com/Gary-Community-Ventures/benefits-api/pull/1004)


src/Types/Results.ts
```diff
@@ -91,11 +91,18 @@ export type UrgentNeed = {
...
   warning: Translation;
   phone_number: string;
+  icon: string;
 };
 
+export type UrgentNeedType = {
+  external_name: string;
+  name: Translation;
+  urgentneeds: UrgentNeed[];
+}
```
CalebPena, line 104: Is this used anywhere?
```suggestion
```

CalebPena (approved):
LGTM

src/Types/Results.ts
msrezaie: Good catch, it is not. Will be removed.

src/Components/Results/Needs/NeedCard.tsx
```diff
@@ -1,17 +1,17 @@
 import { FormattedMessage, useIntl } from 'react-intl';
-import { ComponentType, useState } from 'react';
+import { useState } from 'react';
 import { UrgentNeed } from '../../../Types/Results';
 import ResultsTranslate from '../Translate/Translate';
-import { acuteConditionResultMapping } from '../../../Assets/acuteConditionOptions';
+import { iconCategoryMap } from '../../CurrentBenefits/CurrentBenefits';
```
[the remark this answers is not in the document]
msrezaie: I think it would be best if we do this in a separate ticket. 

Tagging @Linda-Vu in case we create a ticket for this.

PR #1766, opened 2025-06-09
Draft: NC: Feat/add NC LANC

myang1010:
- Add NC LANC program


src/Assets/States/NC/WhiteLabels/LancAssets/lancLink.ts
```diff
@@ -0,0 +1,10 @@
+const LancLinks = [
```
CalebPena, line 1: This is more standard way to do constants.
```suggestion
const LANC_LINKS= [
```

src/Components/LancComponents/LancFooter/LancFooter.css
CalebPena, line 1: I checked this out locally, and it looks like the bottom piece is not centered

image

src/Components/LancComponents/LancHeader/LancHeader.tsx
CalebPena, line 1: A lot of this is duplicated. I think that we should consider breaking some of these pieces out into their own components if we do another one of these. No action for now though.

@jmejia FYI

src/Components/LancComponents/LancShare/LancShare.css
CalebPena, line 1: Messenger has mismatched colors. Also, we recently removed email because it does not work.

image

src/Components/LancComponents/LancShare/LancShare.tsx
CalebPena, line 1: More duplicate stuff here that we might want to consider breaking into its own component or hook. It is less clear for this one on how we might do that though.

src/Components/RouterUtil/RedirectToWhiteLabel.tsx
```diff
@@ -29,6 +29,8 @@ const REFERRER_REDIRECT: { [key: string]: string | undefined } = {
   thegatheringplace: 'co',
   childfirst: 'co',
   eaglecounty: 'co',
+  '211nc': 'nc',
+  lanc: 'nc',
 };
```
CalebPena, line 34: This is only for legacy purposes so that we don't break a couple of old links. Don't add anything to this.

CalebPena (changes requested):
I just realized that this is still a draft, so sorry if this is a premature review.

src/Components/RouterUtil/RedirectToWhiteLabel.tsx
myang1010: I tried redirect without adding them, but the white label gets lost after the redirect. Is there anything else I should do to preserve the white label?

myang1010:
> I just realized that this is still a draft, so sorry if this is a premature review.

Thank you so much for the review, Caleb. It's not premature at all. I'll address them.

src/Components/RouterUtil/RedirectToWhiteLabel.tsx
CalebPena: What redirect are you talking about. We don't want `/step-1?referrer=lanc` to go to `/nc/step-1?referrer=lanc`. We only did that for a couple of old referrer urls because there are existing links pointing to the old spot.

myang1010: I've attached two short videos to help explain the issue. 
In the first video, it includes above lines, and everything works as expected. 

https://github.com/user-attachments/assets/db50d55f-f4ef-4302-8ac5-54bad40f8a45


In the second video, I removed the two lines and when trying to redirect, the "nc" in the url is missing.


https://github.com/user-attachments/assets/50b7a22a-51b7-4118-95cd-28bf821c4ca2

src/Components/LancComponents/LancHeader/LancHeader.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    return dropdownMenuItems;
+  };
+
+  return (
+    <nav>
+      <Paper elevation={4} square={true} className="lanc-header-container">
+        <AppBar position="sticky" id="lanc-nav-container" elevation={0}>
+          <Box>
+            <a href={`/step-1${queryString}`}>
```
CalebPena, line 148: @myang1010 You need to change this link to add the NC. It might also be broken in the 211 white labels, so if you could fix it there as well that would be great.

src/Components/RouterUtil/RedirectToWhiteLabel.tsx
CalebPena: The link is wrong in the header. Lets fix it there instead.

CalebPena (approved):
LGTM

PR #1767, opened 2025-06-09
Central: Confirmation Page Updates

IrynaKolh:
- Central: Confirmation Page Updates

image
image
image
image

https://github.com/user-attachments/assets/6f8839e5-f12d-47c2-87cb-c437ae672250

### DIFF
--- a/tests/nc-end-to-end.spec.ts
+++ b/tests/nc-end-to-end.spec.ts
@@ -35,7 +35,7 @@ test.describe('NC Screen Test', () => {
       await page.getByRole('button', { name: 'Continue' }).click();
     });
     await test.step('Step-5: Household Member 1', async () => {
-      await expect(page).toHaveURL(/\/nc\/.*\/step-5/);
+      await expect(page).toHaveURL(/\/nc\/.*\/step-5\/1$/);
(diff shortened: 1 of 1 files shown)


msrezaie (approved):
lgtm

jmejia:
@sydney-devine UI look OK?

src/Components/Confirmation/ConfirmationSteps.tsx
```diff
@@ -127,7 +127,13 @@ function Expenses() {
 
   const allExpenses = () => {
     if (formData.expenses.length === 0) {
-      return <FormattedMessage id="confirmation.none" defaultMessage="None" />;
+      return (
+        <ConfirmationItem
+          label=""
```
kbrey, line 132: @IrynaKolh can you explain these ConfirmationItem with label=""? It seems based on this code `label &&` that the value won't display if label is an empty string? 
image

kbrey, line 132: @IrynaKolh Small suggestion, it may be clearer to make the label param optional instead of passing an empty string here.

PR #1777, opened 2025-06-13
Improve UX for updating immediate needs

jmejia:
-

QA'ing for release uncovered a few adjustments that needed to be made.

When a user ended up on the Additional Resources/Immediate Needs page, the link to return to edit the immediate needs wasn't editable for each language. Updated the implementation to leverage React Intl's rich text formatting.

The user also would need to step through all remaining steps again after hitting the `Continue` button instead of being returned to the results page. This update returns them back to the page they are coming from.

-
-
-
-

I see two implementations now where we want to route to pages that differ from the standard flow. I wonder if updating our implementation to support something like `redirectTo: <SOME_TARGET>` might be better but I didn't want to mess up state so figured I'd do something I know will work while I'm still getting a feel for the code base. Would love to hear ideas on this if anyone has them as my implementation feels a little code smelly.

### DIFF
--- a/src/Components/QuestionComponents/questionHooks.ts
+++ b/src/Components/QuestionComponents/questionHooks.ts
@@ -7,7 +7,12 @@ import { Context } from '../Wrapper/Wrapper';
 
 export function useShouldRedirectToConfirmation() {
   const location = useLocation();
-  return isCustomTypedLocationState(location.state);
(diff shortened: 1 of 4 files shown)


kbrey (approved):
Looks good to me. I agree with your suggestion to build a more generic "redirectTo" rather than handling each page separately. But it seems to work well and has minimal impact for the release.

PR #1791, opened 2025-06-25
Energy: Use Rewiring America API to get provider list

CalebPena:
- Use RA API to get gas/electric providers instead of hardcoding them.
- Add gas/electric providers to the RA API request.

- There is a tiny load time now on the provider step and a higher chance of an error if the API request fails.


jmejia (approved):
A couple questions which has me wondering if there's an opportunity to DRY up some of the repetition but I trust your judgement.

Issue #1831, opened 2025-07-21
Playwright E2E Failing Test


mexi-cano:
<img width="1509" height="969" alt="Image" src="https://github.com/user-attachments/assets/432aaa9f-4c65-421a-9b3b-39bce526801a" />

PR #1832, opened 2025-07-21
Fix NC E2E Failing Test #1831

mexi-cano:
- Fixed NC end-to-end tests failing due to "element detached from DOM" 
errors when selecting birth dates in the household member step.


IrynaKolh (approved):
LGTM

Issue #1677, opened 2025-05-09
Central: Create Playwright tests around error messages


Linda-Vu:
Context:

Write E2E test NC 211 cobrand
Test case -  https://docs.google.com/document/d/1hjYl60K786BXeZoPIqoqQDRCToaOUnr2PkLGBeZVaXU/edit?tab=t.0#heading=h.hel7ikg1a2dg

Dev details:

PR #1853, opened 2025-07-28
Central: Playwright tests around error messages

myang1010:
#1677 
- Add Playwright tests to verify error messages at each step


kbrey (approved):
Looks excellent @myang1010 !

PR #1867, opened 2025-07-31
Energy: Add support for renters RA rebates

jmejia:
-

Rewiring America supports renters rebates through their API. We initially were hardcoding `homeowner` in our API calls. This PR sends the `homeowner` or `renter` option dynamically and supports results that are specific to both.

Most of the renter-supported rebates are nested under `Efficiency & Weatherization` in the RA API so we've also added a new category to organize these rebates.

-
-
-
-


patwey (approved):
Nice, LGTM!

PR #1872, opened 2025-07-31
Energy: Add Landlord's Consent Statement

jmejia:
-

We initially only offered the rebates to homeowners. Now that the results include renters, we're updating the language to include a statement about needing the landlord's consent.

-
-
-
-


patwey (approved):
You have my consent for these changes!

Issue #1877, opened 2025-08-05
Energy: Update favicon


jmejia:
Solution should allow other white labels to provide custom favicons

PR #1881, opened 2025-08-05
Central/Energy: Add support for custom favicon

jmejia:
-

https://github.com/Gary-Community-Ventures/benefits-calculator/issues/1877

-
-
-
-


patwey (approved):
Looks good to me!

Out of curiosity - do we need to do any testing with older browsers (e.g. via something like BrowserStack) to verify the browser support built into `FaviconManager`?

jmejia:
> Looks good to me!
> 
> Out of curiosity - do we need to do any testing with older browsers (e.g. via something like BrowserStack) to verify the browser support built into `FaviconManager`?

@patwey Great question. As far as I'm aware, we haven't been doing this but it doesn't mean we shouldn't. Might be worth revisiting as an engineering team and adding to our CI pipeline in the near future. Great callout.

PR #1889, opened 2025-08-08
Feat/Cobrand-Referrers default language

msrezaie:
- Added a `useEffect` hook to update the default language of the screener only for `referrers` that has a default language set using a new `defaultLanguage` object in its whitelabel's configuration record.

- Please follow the setup steps in this PR [#1097](https://github.com/Gary-Community-Ventures/benefits-api/pull/1097) to update the configuration records.

### DIFF
--- a/src/Components/Wrapper/Wrapper.tsx
+++ b/src/Components/Wrapper/Wrapper.tsx
@@ -189,6 +189,17 @@ const Wrapper = (props: PropsWithChildren<{}>) => {
     document.documentElement.removeAttribute('dir');
   }, [locale]);
 
+  useEffect(() => {
+    if (!referrerData) return;
+    const defaultLang = getReferrer('defaultLanguage', 'en-us');
+    const theme = getReferrer('theme', 'default');
+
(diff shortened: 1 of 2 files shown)


src/Components/Wrapper/Wrapper.tsx
```diff
@@ -189,6 +189,17 @@ const Wrapper = (props: PropsWithChildren<{}>) => {
     document.documentElement.removeAttribute('dir');
   }, [locale]);
 
+  useEffect(() => {
+    if (!referrerData) return;
+    const defaultLang = getReferrer('defaultLanguage', 'en-us');
+    const theme = getReferrer('theme', 'default');
+
+    const isCobrand = theme !== 'default';
```
kbrey, line 197: I'm wondering about this `isCobrand` check, I think we might want to remove it. My thinking is, in the hispanic federation case (referrer code is hfed), I think this condition would prevent their co-brand from defaulting to Spanish like we want. I think maybe the term "cobrand" can be interpreted multiple ways but in my mind it's any referrer code we use.

PR #2096, opened 2026-04-07
Use screener-options endpoint for referral sources

catonph:
## Context & Motivation

Previously, referral options were sourced from static white label config files baked into the Django config system. This PR switches the FE to fetch referral options dynamically from the new `/screener-options/{wl}/referral-options/` API endpoint (introduced in benefits-api#1448), which reads directly from the `programs_referrer` Django model.

- Closes MFB-864 (benefits-calculator)
- Related PRs: MyFriendBen/benefits-api#1448, MyFriendBen/data-queries#68
- Follow-up: #2098 (conditional step skip logic, depends on this PR)

## Changes Made

**`src/hooks/useReferralOptions.ts`** (renamed from `useReferralSources`)
- Fetch referral options from the new API endpoint
- Return grouped `{ generic, partners }` response matching the new API shape

**`src/apiCalls.ts`**
- Add `getReferralOptions` function (renamed from `getReferralSources`)

**`src/Components/Steps/Referrer.tsx`**
- Update to use grouped `referralOptions` response
- Render partners and generic options in separate sections with subheaders and dividers
- Alphabetical sort within each group

**`src/Components/Confirmation/ConfirmationSteps.tsx`**
- Update guard for referralSource step to use new grouped options shape
- Fix: custom "other" text (typed by user) now displays verbatim on confirmation page instead of being hidden

## Testing

- Migrations to run: `python manage.py migrate programs 0142` (seeds `programs_referrer` from existing referral_options config)
- Configuration updates needed: `python manage.py add_config --all` (updates referral_options config rows ahead of migration)
- Environment variables/settings to add: none
- Manual testing steps:
  - [ ] Visit `/co/` with no referrer param — referral source step appears, dropdown shows General and Partners sections with alphabetical sort within each group
  - [ ] Verify General section contains: Flyer, Friend / Family / Word of Mouth, Google or other search engine, Merit America, Other, Social Media
  - [ ] Verify Partners section contains the expected CO partner orgs (2-1-1 Colorado, Salud, etc.)
  - [ ] Simulate API failure (block the referral-options request in devtools) — referral source step still appears, no crash
  - [ ] **Custom "other" text on confirmation page:** complete the screener, select "Other" as referral source, type a custom value (e.g. "local community center"), continue through to the confirmation page — verify the confirmation page shows "local community center" verbatim in the Referral Source block (regression: this was hidden before this fix)

## Deployment

- Run migrations: `python manage.py migrate programs 0142`
[...]

PR #2098, opened 2026-04-09
MFB-876: Only skip referralSource step for recognized referrer codes

catonph:
## Context & Motivation

- Closes MFB-876
- **Depends on #2096 being merged first**

MFB-859 added logic to skip the \`referralSource\` step whenever a \`referrer\` query param is present. Without this fix, any unrecognized referrer code silently skips the step, leaving \`referral_source\` set to the raw code with no user input.

This PR tightens that logic: only skip the step when the referrer code is actually recognized (exists in the referral options from the API).

## Changes Made

**\`src/Components/Wrapper/Wrapper.tsx\`**
- Lift referral options fetch into global context so \`useStepDirectory\` can access it synchronously
- Fetches once per white label on app load

**\`src/Types/WrapperContext.ts\`**
- Add \`referralOptions\` and \`referralOptionsLoading\` to context interface

**\`src/Assets/stepDirectory.ts\`**
- Skip \`referralSource\` step only if \`immutableReferrer\` is set AND is a recognized code
- While options are loading, show the step (safe default — no flicker in practice since fetch starts at app load, well before the user reaches this step)

**\`src/Assets/stepDirectory.test.tsx\`** (new)
- 6 tests covering all scenarios

**Test mocks updated:** \`renderHelpers.tsx\`, \`configHook.test.tsx\`, \`ProgramCard.test.tsx\`, \`Filter.test.tsx\`

## Behavior

| Scenario | Result |
|---|---|
| No referrer param | Step shown |
| Known referrer code (e.g. \`?referrer=dhs\`) | Step skipped |
| Unknown referrer code (e.g. \`?referrer=unknowncode\`) | Step shown, "other" selected, code pre-filled |
| Options still loading | Step shown (safe default) |
| Options failed to load | Step shown (safe default) |

## Manual Testing

- [ ] Visit \`/co/\` with \`?referrer=dhs\` — referral source step should be skipped
- [ ] Visit \`/co/\` with \`?referrer=unknowncode\` — referral source step appears with "other" selected and "unknowncode" pre-filled
- [ ] Visit \`/co/\` with no referrer param — referral source step appears normally


catonph (commented):
LGTM!

cdadams1888 (approved):

PR #2118, opened 2026-06-01
MFB-921: Adding has_wap field

david-j-stoll:
## Context & Motivation

- Fixes #MFB-921 (WA Weatherization Assistance Program)
- The WA white label was missing frontend support for the Weatherization Assistance Program (WAP). The backend already exposes `has_wap` on the Screen API, but the frontend was not reading from or writing to that field, meaning WAP enrollment status could not be captured or round-tripped through the screener.

## Changes Made

- `src/Types/ApiFormData.ts`: Added `has_wap: boolean | null` to the `ApiFormData` type so the API response shape is correctly typed.
- `src/Assets/updateFormData.tsx`: Added `wa_wap: response.has_wap ?? false` to map the `has_wap` API field into the frontend form state, following the same pattern used for `cowap` and `ncwap`.
- `src/Assets/updateScreen.ts`: Added `has_wap: formData.benefits.wa_wap ?? null` so that the `wa_wap` benefit selection is sent back to the API when the screen is updated.

## Testing

- Migrations to run: None (frontend-only change; `has_wap` already exists on the backend model).
- Configuration updates needed: Ensure the WA white label has a `wa_wap` benefit configured in the step 8 fan-out on the backend.
- Environment variables/settings to add: None.
- Manual testing steps:
  1. Open the screener on the WA white label.
  2. Navigate to step 8 (existing benefits) and select the WA Weatherization Assistance Program.
  3. Submit the step — verify the API request body contains `"has_wap": true`.
  4. Reload the screen — verify the WAP checkbox is pre-populated from `response.has_wap`.

## Deployment

- Run script: None.
- Update production config: Confirm the WA white label step 8 fan-out includes `wa_wap` on the backend (backend PR may be a prerequisite or companion).
- Admin updates needed: None.
- Notify team/users of: WA WAP benefit enrollment can now be captured and stored through the screener.

## Notes for Reviewers

- Unlike `wa_snap` and `wa_wic`, `wa_wap` maps to its own dedicated backend field (`has_wap`) rather than sharing a field with other state variants. `cowap` and `ncwap` each have their own fields (`has_cowap`, `has_ncwap`) — `has_wap` follows this same per-state pattern.
- `wa_wap` is not currently included in a sibling group in `benefitSiblings.ts`. If WAP from another state is ever added with a shared `has_wap` field, a sibling group may be needed.
- The default benefit state in `Wrapper.tsx` does not yet include `wa_wap: false`. This matches how other WA-specific benefits (`wa_snap`, `wa_wic`, `wa_hcv`) are handled — they rely on the step 8 fan-out configuration rather than a hardcoded default.


catonph (approved):
LGTM

PR #2124, opened 2026-06-09
MFB-1027: Add new resource buttons and resources from New Button Resources spreadsheet

david-j-stoll:
## Context & Motivation

- Implements [MFB-1027](https://linear.app/myfriendben/issue/MFB-1027/add-new-resource-buttons-and-resources-from-new-button-resources) — "Add new resource buttons and resources from New Button Resources spreadsheet".
- The Additional Resources step (acute conditions) renders selectable category buttons from the backend `acute_condition_options` config. A new **Aging Community Resources** category is being introduced (and the existing Disability category extended to more states), so the frontend needs to render the new button's icon and round-trip its selection through the screen API.
- Also unifies the **Disability** category icon: the result-page card now uses the same disability glyph as the selection button (previously the disability result card reused the generic `savings`/Resources icon).
- Requires the corresponding backend changes (new `agingResources` config option + `needs_aging_resources` screen field; disability `category_type.icon` changed to `disability`). See the benefits-api PR.

## Changes Made

- **New icon asset:** `src/Assets/icons/UrgentNeeds/AcuteConditions/aging.svg` — the Aging Community Resources icon. Stripped from the raw svgrepo export to the icon paths (`class="cls-1"`, no fixed `width`/`height`), with `stroke-width="1.52"` + round line caps/joins on the root so it renders as a thin outline. Used in **two** places: the Additional Resources button and the results-page card.
- **Button icon wiring:** `src/Components/Config/configHook.tsx` — imported the `Aging` SVG component and added a `case 'Aging'` to `transformItemIcon`, mapping the backend `_icon: "Aging"` token to the asset. The icon is rendered with an extra `aging-icon` class (`className={`${icon._classname} aging-icon`}`) so it can be styled as an outline.
- **Results-card icon wiring (aging):**
  - `src/Components/Results/helpers.ts` — imported the `Aging` SVG and registered `aging: Aging` in `ICON_OPTIONS_MAP` (a new key; backend resources set `category_type.icon: "aging"`).
  - `src/Components/Results/Needs/NeedCard.tsx` — `getIcon` now passes an `aging-icon` class to the rendered icon only when the resource's icon is `aging`; all other card icons are unchanged.
- **Results-card icon wiring (disability):** `src/Components/Results/helpers.ts` — imported the existing `disabled.svg` (the **same asset** the selection button uses via `configHook`'s `case 'Disabled'`) and registered `disability: Disabled` in `ICON_OPTIONS_MAP`. Backend disability resources now set `category_type.icon: "disability"` (previously `savings`), so the result card shows the same disability glyph as the button. **No new asset and no CSS** needed: `disabled.svg` is a solid-fill `path.cls-1` icon already handled by the base `.result-resource-more-info svg path.cls-1 { fill: var(--icon-color) }` rule (unlike aging, which is line art and needs the outline overrides below).
- **Outline styling (aging — orange line art instead of solid fill):**
  - `src/Components/SelectTiles/SelectTiles.css` — added `.option-card-icon.aging-icon * { fill: none !important; stroke: var(--icon-color) !important }` (overrides the force-fill in `.option-card-icon *`) plus a selected-state override (`fill: none; stroke: white`).
  - `src/Components/Results/Needs/NeedCard.css` — added `.result-resource-more-info svg.aging-icon { fill: none; stroke: var(--icon-color) }` (higher specificity than the base fill rule, so no `!important` needed).
  - Stroke width is driven by the SVG's `stroke-width="1.52"`; since both contexts render the icon at `4.5rem` over the same `57.6` viewBox, the on-screen line weight is identical in both. Changing that one value updates both places.
- **API type:** `src/Types/ApiFormData.ts` — added `needs_aging_resources: boolean | null` to the screen payload type.
- **Form → API mapping:** `src/Assets/updateScreen.ts` — maps `formData.acuteHHConditions.agingResources` to `needs_aging_resources` when saving a screen.
- **API → Form mapping:** `src/Assets/updateFormData.tsx` — maps `response.needs_aging_resources` back to `acuteHHConditions.agingResources` when loading a screen.

## Testing

- **Migrations to run:** None (frontend only). Requires the benefits-api branch with the `needs_aging_resources` field/migration and the updated `acute_condition_options` config applied to the API the frontend points at.
- **Configuration updates needed:** Point the app at a backend where `add_config` has pushed the new `agingResources` (and IL/TX `disabilityResources`) buttons for MA/TX/WA/IL, and where the disability resources have been (re-)imported with `category_type.icon: "disability"`; otherwise the new button won't appear / the result card will still show the old icon.
- **Environment variables/settings to add:** None.
- **Manual testing steps:**
  1. `npm install` (if needed) and `npm start`.
  2. For each of MA/TX/WA/IL, navigate to the Additional Resources step and confirm the **Aging Community Resources** tile renders the aging icon as a thin **orange outline** (not a solid fill), and that it inverts correctly in the selected state (white outline on the selected tile).
  3. Confirm IL and TX also show **Resources for people with disabilities** (and MA still does).
  4. Select the aging tile, continue, and confirm the selection persists (reload / navigate back) — verifying `needs_aging_resources` round-trips through save/load.
  5. Confirm the selected aging resources appear on the results page, and that the aging card icon renders as the same orange outline (other card icons remain solid fills).
  6. Select the disability category (IL/TX/MA), and confirm the disability resource appears on the results page with the **same disability icon as the selection button** (not the old `savings`/Resources icon).
  7. `npx tsc --noEmit` — no new type errors introduced by these files (pre-existing repo errors are unrelated).

## Deployment

- **Run script:** Standard frontend build/deploy (`npm run build` → deploy) for the environment(s).
- **Update production config:** No frontend config change. The backend must have run `add_config` (new buttons) so the config endpoint serves `agingResources`/`disabilityResources`, and must have (re-)imported the disability resources with the `disability` icon; otherwise the tile/icon won't show or the result card will keep the old icon.
- **Admin updates needed:** Ensure the translation for `acuteConditionOptions.agingResources` exists in the Translation table for prod (the `_default_message` provides a fallback, but non-English locales should be verified).
- **Notify team/users of:** New Aging Community Resources category live in MA, TX, WA, IL; the new disability resources in IL and TX; and the unified disability result-card icon.

## Notes for Reviewers

- **Known limitations:**
  - The Additional Resources step renders options dynamically from config, so no per-option component change was needed — only the icon mapping and the screen-field round-trip.
  - The `Wrapper` default `acuteHHConditions` and the `ApiAcuteHHConditions` type in `Config.ts` were intentionally left unchanged: they already omit state-specific options like `veteranServices`/`disabilityResources`, which work fine, so `agingResources` follows the same pattern.
  - The disability result-card change reuses the **existing** `disabled.svg` (no new asset, no CSS) — it just adds a `disability` key to `ICON_OPTIONS_MAP` pointing at the same icon the button already uses. It only takes effect once the backend serves `category_type.icon: "disability"` (see the benefits-api PR's re-import note for the pre-existing `ma_able_account` resource).
  - The provided aging source SVG was a raw svgrepo export (fixed dimensions, wrapper groups, hardcoded colors); it was stripped down to the icon paths to match the directory convention.
  - The aging icon is the **only** outline-style icon in that directory — every other acute-condition/result-card icon is a solid filled silhouette. The outline is achieved via the dedicated `aging-icon` class + CSS (mirroring the existing `food-icon`/`youth-development-icon` per-icon class pattern) rather than by changing the SVG fill, because that same SVG is reused on the results card (a `fill: none` in the file would make it disappear where no stroke is applied).
  - **No unit tests added.** `benefits-calculator/tests/` is the Playwright E2E suite (no coverage of these cards), and there is no existing `NeedCard.test.tsx` / `configHook` icon-switch test to extend. Adding `NeedCard`/icon-mapping unit tests is a reasonable follow-up.
- **Future considerations:**
  - If `ApiAcuteHHConditions` is ever made authoritative, it should be updated to include `veteranServices`, `disabilityResources`, and `agingResources`.
  - The `transformItemIcon` switch keeps growing per icon; a lookup map could reduce boilerplate if more options are added.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking "Aging Resources" needs in user forms and request payloads.
  * Integrated a new aging resources icon for consistent display across configuration and results.


src/Assets/icons/UrgentNeeds/AcuteConditions/aging.svg
catonph, line 1: Can we do the outline version? https://www.svgrepo.com/svg/325902/elderly

src/Components/Results/Needs/NeedCard.css
```diff
@@ -19,6 +19,14 @@
...
 }
 
+/* Render the aging icon as an outline (line art) instead of a solid fill;
+   stroke-width comes from the SVG. Matches the Additional Resources button. */
+.result-resource-more-info svg.aging-icon,
+.result-resource-more-info svg.aging-icon path.cls-1 {
+  fill: none;
+  stroke: var(--icon-color);
+}
+
```
catonph: Probably don't need to do this if you use https://www.svgrepo.com/svg/325902/elderly?

src/Components/Results/Needs/NeedCard.tsx
```diff
@@ -15,7 +15,7 @@ const getIcon = (messageType: string) => {
   const Icon = ICON_OPTIONS_MAP[messageType] ?? ICON_OPTIONS_MAP['default'];
 
   if (Icon !== undefined) {
-    return <Icon />;
+    return <Icon className={messageType === 'aging' ? 'aging-icon' : undefined} />;
```
catonph: Why are we suddenly doing a className here just for aging? This feels like the wrong place for this?

catonph: Should be able to apply a classname in src/Components/Config/configHook.tsx

src/Components/SelectTiles/SelectTiles.css
```diff
@@ -175,6 +175,13 @@
   stroke-width: 0.313rem;
 }
 
+/* Render the aging icon as an outline (line art) instead of a solid fill.
+   Overrides the force-fill in `.option-card-icon *`; stroke-width comes from the SVG. */
+.option-card-icon.aging-icon * {
+  fill: none !important;
+  stroke: var(--icon-color) !important;
+}
```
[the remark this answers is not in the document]
catonph: Worth checking this?

src/Components/SelectTiles/SelectTiles.css
```diff
@@ -184,6 +191,12 @@
   stroke: white !important;
 }
 
+/* Keep the aging icon an outline when selected (white line on the selected fill) */
+.option-card--selected .option-card-icon.aging-icon * {
+  fill: none !important;
+  stroke: white !important;
+}
+
```
catonph: Maybe not needed if you start with an outlined one

catonph (changes requested):
A few requests/suggestions

catonph (approved):

PR #2122, opened 2026-06-03
MFB-981: Heat Pump Step 3 - Wire Rewiring America API + EPA equivalencies for heat pump water heater

david-j-stoll:
## Context & Motivation

- Related to Linear ticket MFB-981 (Heat Pump Step 3: Wire Rewiring America API + EPA equivalencies for heat pump water heater)
- The CESN Calculate Impact page (built in MFB-979) had a working form UI but a no-op submit handler. This PR wires the frontend half of the integration — calling the new backend REM proxy endpoint (added in the companion `benefits-api` PR), rendering the results view, and displaying EPA carbon equivalencies.
- All new UI is gated behind the existing `cesn_heat_pump_journey` feature flag.

## Changes Made

- **`src/utils/epaEquivalencies.ts`** (new): Pure TypeScript utility that converts lb CO₂e to EPA equivalency values. Currently implements one equivalency — acres of U.S. forests sequestering carbon for one year — using the EPA's published factor of 1.00 metric ton CO₂/acre/yr (source cited in file). Future equivalencies (miles driven, gallons of gasoline, etc.) are stubbed as comments and can be enabled by appending entries to the `EPA_EQUIVALENCIES` array.
- **`src/utils/epaEquivalencies.test.ts`** (new): 6 unit tests covering the forest acres conversion — correct value for 1 metric ton, absolute value behavior, zero, linear scaling, and message ID.
- **`src/Components/EnergyCalculator/Results/HeatPumpJourney/remCalculateImpactTypes.ts`** (modified): Added `RemImpactStat`, `RemImpactRange`, and `RemImpactApiResponse` interfaces matching the backend endpoint's response shape. Added `CalculateImpactFormValues` interface (shared between the form and results summary). Added `HOUSEHOLD_TYPE_LABEL_MAP`, `FUEL_TYPE_LABEL_MAP`, and `UPGRADE_LABEL_MAP` lookup objects used by the results summary card.
- **`src/Components/EnergyCalculator/Results/HeatPumpJourney/fetchRemImpact.ts`** (new): Async function that calls `GET /api/screener-options/<white_label>/rem-impact/` with the form payload and returns a typed `RemImpactApiResponse`.
- **`src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactResults.tsx`** (new): Full results view rendered after a successful API call. Contains three sections:
[...]

Issue #1559, opened 2025-03-10
Update the names of labels in nc.py


Linda-Vu:
Address differences in configurable content by updating the names of labels in nc.py and creating new translations manually

Linda-Vu:
may not need this ticket, tbd

PR #2126, opened 2026-06-11
MFB-982: Extend "Calculate impact" calculator to remaining upgrade types

david-j-stoll:
# MFB-982: "Calculate Impact" calculator — offer heat pump & heat pump water heater upgrades

Limit the CESN "Calculate Impact" form to the two upgrade types whose bill impact REM can model reliably — **heat pump** and **heat pump water heater**. Weatherization-based upgrades are intentionally **not** offered as selectable options; per CESN SME guidance their bill impact can't be modeled without knowing the home's current condition. Their potential benefit is surfaced as a static note on the results page instead.

> **Note on scope change:** An earlier revision of this branch un-gated all four upgrade types (heat pump, weatherization, heat pump + weatherization, HPWH). That approach was reversed after SME guidance — this PR removes the two weatherization options rather than enabling them. This description reflects the current diff.

## Context & Motivation

- Implements Linear ticket **MFB-982** (Heat Pump Step 4: extend the "Calculate impact" calculator beyond the single HPWH option that shipped in Step 3).
- Builds on the now-merged Step 3 work: `benefits-api` #1559 (REM proxy endpoint) and `benefits-calculator` #2122 (form, results view, EPA equivalencies). Step 3 shipped the calculator end-to-end but enabled only the **heat pump water heater (HPWH)** upgrade; the other three options rendered disabled with a "Coming soon" badge.
- This PR enables the **heat pump** upgrade alongside HPWH and removes the weatherization-based options entirely (rather than enabling them), since their savings can't be modeled reliably. The "Coming soon" gating is gone.
- **Backend:** no changes required. The `rem-impact` proxy forwards `upgrade` as an opaque string and the serializer extracts `fuel_results.total.delta` generically.

## Changes Made

All changes are in `src/Components/EnergyCalculator/Results/HeatPumpJourney/`.

- **`remCalculateImpactTypes.ts`** — Narrowed `CalculateImpactUpgradeChoice` to `'heat_pump' | 'heat_pump_water_heater'` and removed the `weatherization` / `heat_pump_weatherization` entries from `UPGRADE_LABEL_MAP` and `CALCULATE_IMPACT_UPGRADE_MAP`. The type is now the source of truth — no weatherization choice can be constructed. The two weatherization `RemUpgrade` API enum strings are retained (with a comment) because they remain valid REM API inputs.
- **`CalculateImpactPage.tsx`** — `UPGRADE_OPTIONS` / `UPGRADE_CHOICE_VALUES` now list only heat pump and HPWH; the "Coming soon" badge and the per-option `data-disabled` flag are gone. Reworded the `heatPump.description` and `heatPumpWaterHeater.description` defaultMessages for clarity (same message IDs — see Translations).
- **`CalculateImpactResults.tsx`** — The "…adding air conditioning…" bill disclaimer (`energyCalculator.calculateImpact.results.billImpact.however`) now renders only for `heat_pump` (the whole-home space-conditioning upgrade), not for HPWH. Added a static weatherization note (`energyCalculator.calculateImpact.results.weatherizationNote`, **new copy**) explaining why weatherization savings aren't included in the estimate.
- **`CalculateImpactPage.css`** — Removed the now-orphaned `.calculate-impact-coming-soon` and `[data-disabled='true']` rules.
- **Tests** — Updated `CalculateImpactPage.test.tsx` and `CalculateImpactResults.test.tsx` to assert the two-option set, absence of weatherization options / "Coming soon", the conditional AC disclaimer, and the weatherization note. `remCalculateImpactTypes.test.ts` now asserts the two-choice map.

## Testing

- **Migrations to run:** None.
- **Configuration updates needed:** Ensure the `cesn_heat_pump_journey` feature flag is enabled in the white label config for the environment under test.
- **Manual testing steps:**
  1. Start backend and frontend dev servers.
  2. Navigate to `/{whiteLabel}/{uuid}/results/energy-rebates/waterHeater/calculate-impact`.
  3. Confirm exactly **two** upgrade options (heat pump, heat pump water heater) are selectable and **no** "Coming soon" badge or weatherization options appear.
  4. Submit each upgrade type and confirm the results view renders sensible bill + emissions range bars and EPA equivalencies.
     - For **heat pump**, water-heating fuel can be left blank.
     - For **heat pump water heater**, water-heating fuel is required (validation error if blank).
  5. Confirm the "…adding air conditioning…" disclaimer shows **only** for **heat pump**, and is **absent** for **heat pump water heater**.
  6. Confirm the weatherization note renders on the results page.
  7. Run unit tests: `npx react-scripts test --env=jsdom --watchAll=false --testPathPattern="HeatPumpJourney"` (72 tests pass).

## Deployment

- **Run script:** None.
- **Update production config:** None required by this PR. The journey remains gated behind `cesn_heat_pump_journey` (still `false` in production); flipping that flag is a separate, deliberate step.
- **Admin / Translation updates needed:** **Yes — must be done in each environment (local → staging → prod) BEFORE flipping `cesn_heat_pump_journey` on.** The full rationale is in the Translations section below; the concrete steps are:

  1. **Reworded descriptions (existing message IDs — manual admin-portal edit, NOT `add_translations`).** In the Django admin `Translation` records, update both the English row and every non-English locale (es, etc.) for:
     - `energyCalculator.calculateImpact.upgrade.heatPump.description`
       > A heat pump is an appliance that works in two modes, heating and cooling. By moving heat from the surrounding air (instead of creating it), heat pumps use less energy, making them more efficient than traditional systems.
     - `energyCalculator.calculateImpact.upgrade.heatPumpWaterHeater.description`
       > A heat pump water heater moves heat from the surrounding air to the water tank to heat the water. Because of this, it uses significantly less energy than traditional electric or gas water heaters.

     Because the message IDs already exist, do **not** run `add_translations` for these — it would re-translate over the IDs. Edit the rows directly so non-English locales get the new wording (English `defaultMessage` in code is only a fallback).
  2. **New weatherization note (new message ID — standard `add_translations` flow).** Run the normal `add_translations` flow on:
     - `energyCalculator.calculateImpact.results.weatherizationNote`
       > Weatherization — like insulation and air sealing — can also lower your energy bills and improve comfort. Because the savings depend heavily on your home's current condition, they aren't included in this estimate.

     No `{placeholder}` tokens in any of these three strings, so no ICU mangling repair is needed afterward.
  3. **Orphaned IDs cleanup** The removed `upgrade.weatherization*` / `upgrade.heatPumpWeatherization*` and `comingSoon` IDs are no longer rendered; their Translation rows are harmless but SHOULD be removed in the admin section

## Translations

This PR **does** change copy. Two things need attention in the backend `Translation` table:

1. **New message ID — `energyCalculator.calculateImpact.results.weatherizationNote`** (the static weatherization note). This is genuinely new copy and needs the standard `add_translations` flow so non-English locales get translated rows. Repair any ICU `{placeholder}` mangling in the auto-translated rows afterward (none expected here — the string has no placeholders).
2. **Reworded defaultMessages, same IDs — non-English rows are now stale:**
   - `energyCalculator.calculateImpact.upgrade.heatPump.description`
   - `energyCalculator.calculateImpact.upgrade.heatPumpWaterHeater.description`
   English now shows the new wording (the React `defaultMessage`), but every other locale keeps the old text until the rows are updated. Because the IDs already exist, this is **not** an `add_translations` run — it's a manual edit per environment in the admin portal. Do this before the flag flips on.

The removed `weatherization` / `heatPumpWeatherization` upgrade label/description IDs and the `comingSoon` ID are no longer rendered; their Translation rows are now orphaned (harmless) but SHOULD be removed in the admin section

## Notes for Reviewers

- The type narrowing makes `CalculateImpactUpgradeChoice` the single gate for which upgrades can exist, rather than relying on `UPGRADE_CHOICE_VALUES` / `UPGRADE_OPTIONS` arrays alone.
- Weatherization `RemUpgrade` API strings are intentionally retained — they're valid REM inputs even though no app-level choice maps to them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -126,10 +110,11 @@ const HOUSEHOLD_TYPE_VALUES: [CalculateImpactHouseholdType, ...CalculateImpactHo
...
 
+// Weatherization-based upgrades are intentionally excluded: per CESN SME guidance,
+// their bill impact can't be modeled reliably without knowing the home's current
+// condition. The potential impact is surfaced as a note in the results instead.
 const UPGRADE_CHOICE_VALUES: [CalculateImpactUpgradeChoice, ...CalculateImpactUpgradeChoice[]] = [
   'heat_pump',
-  'weatherization',
-  'heat_pump_weatherization',
   'heat_pump_water_heater',
 ];
```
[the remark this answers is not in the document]
david-j-stoll: I think I want to keep these in in case they change their mind. It'll make those additions a small PR, and seems like it's still bound to happen in the near future.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactResults.tsx
```diff
@@ -169,6 +169,10 @@ export default function CalculateImpactResults({
     : null;
   const upgradeLabel = UPGRADE_LABEL_MAP[formValues.upgradeChoice];
 
+  // The AC disclaimer only applies to upgrades that add a heat pump for space
+  // conditioning (whole-home heat pump); it's misleading for the water-heater upgrade.
+  const showAcDisclaimer = formValues.upgradeChoice === 'heat_pump';
+
```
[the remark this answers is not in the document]
david-j-stoll: There is no weatherization option at this time.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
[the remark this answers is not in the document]
catonph: I think I'd vote for removing these options to avoid confusion. The email from Debra was 

> We checked with our weatherization subject-matter experts and they would like to remove the weatherization options in the calculator. Here is what a program lead said: "While I love the idea of putting weatherization on folk's radar, I have concerns because bill impacts may be difficult calculate unless someone understands the their home's current condition as it relates to energy efficiency measures." 

It doesn't really feel like they'll come back any time soon with wanting to add them back in?

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -89,31 +89,15 @@ const UPGRADE_OPTIONS: {
...
-      'Weatherization includes insulation, air sealing, and other improvements that reduce energy waste and make your home more comfortable year-round.',
-  },
-  {
-    value: 'heat_pump_weatherization',
-    messageId: 'energyCalculator.calculateImpact.upgrade.heatPumpWeatherization',
-    defaultMessage: 'Heat pump + weatherization',
-    descriptionId: 'energyCalculator.calculateImpact.upgrade.heatPumpWeatherization.description',
-    descriptionDefault:
-      'Combining a heat pump with weatherization maximizes energy savings by pairing efficient heating and cooling with a well-sealed, insulated home.',
+      'A heat pump is an appliance that works in two modes, heating and cooling. By moving heat from the surrounding air (instead of creating it), heat pumps use less energy, making them more efficient than traditional systems.',
```
catonph, line 92: Reworded `defaultMessage` without changing the message ID. Because the non-English text lives in the backend `Translation` table, the existing rows for `upgrade.heatPump.description` are now out of sync — English shows this new wording, every other locale keeps the old text until someone updates those rows. Per our usual flow that's a manual edit in the admin portal (per environment), not an `add_translations` run. Worth doing before the flag flips on, and worth noting in the PR's Translations section since it currently says no copy changed. Same applies to the `heatPumpWaterHeater.description` change below.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactResults.tsx
```diff
@@ -169,6 +169,10 @@ export default function CalculateImpactResults({
     : null;
   const upgradeLabel = UPGRADE_LABEL_MAP[formValues.upgradeChoice];
 
+  // The AC disclaimer only applies to upgrades that add a heat pump for space
+  // conditioning (whole-home heat pump); it's misleading for the water-heater upgrade.
+  const showAcDisclaimer = formValues.upgradeChoice === 'heat_pump';
```
catonph, line 174: Since the recommendation is to drop weatherization entirely, this `=== 'heat_pump'` check is fine — just make sure the `(whole-home heat pump)` comment still reads correctly once the `heat_pump_weatherization` member is gone, so nobody later wonders why that case isn't handled.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -594,29 +579,19 @@ export default function CalculateImpactPage() {
...
                   {UPGRADE_OPTIONS.map((opt) => {
                     const isSelected = field.value === opt.value;
-                    const isAvailable = opt.value === 'heat_pump_water_heater';
                     return (
                       <div
                         key={opt.value}
                         className="calculate-impact-radio-option"
                         data-selected={isSelected ? 'true' : 'false'}
-                        data-disabled={!isAvailable ? 'true' : 'false'}
+                        data-disabled="false"
```
catonph: `data-disabled` is now a permanently-`"false"` constant on every option. If nothing is ever disabled anymore, drop the attribute and any `[data-disabled='true']` CSS rules rather than leaving a hardcoded-false flag. Minor.

catonph (commented):
Code is clean and the tests are reasonable, but two things need attention before this can go live:

**The PR description doesn't match the diff.** The description says this PR *un-gates all four upgrade types* and introduces *no new copy*. The diff does the opposite — it **removes** the two weatherization options (`weatherization`, `heat_pump_weatherization`), keeps only `heat_pump` + `heat_pump_water_heater`, and **adds** a new static weatherization note plus reworded descriptions. Please rewrite the PR body so reviewers (and the changelog) describe the actual change.

**Dead code left behind — confirm intentional.** The `comingSoon` message ID + CSS, and the `weatherization` / `heat_pump_weatherization` entries in `UPGRADE_LABEL_MAP` / `CALCULATE_IMPACT_UPGRADE_MAP` (`remCalculateImpactTypes.ts`), are now unreachable from the UI but still in the `CalculateImpactUpgradeChoice` union. Keeping the type members is fine if weatherization may return, but worth a comment so the next dev doesn't assume that path is live.

Inline comments below for the must-fix translation items and a couple of nits.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
david-j-stoll: I called this out in the description to run in staging and prod and ran locally, good catch

catonph (approved):
LGTM!

PR #2145, opened 2026-07-14
Instrument CESN heat pump journey events [MFB-1582]

catonph:
Frontend instrumentation for the CESN heat-pump / HVAC results-page journey. Emits the typed `heat_pump_*` events through `useTrackEvent` (same pipeline as `screener_*`), so they land in the GA4 `events_*` export with `screener_uid` attached. Feeds the CESN dashboard tab in data-queries PR #129.

## Events

**Clicks (numerators):**
- `heat_pump_cta_click { cta: calculate_impact | connect_now }` — the two internal CTAs
- `heat_pump_calculator_field { field }` — each calculator field engaged
- `heat_pump_calculator_submit { household_type?, heating_fuel?, water_heating?, project_type? }`
- `heat_pump_calculator_edit` — edit after results
- `heat_pump_calculator_error { error_type: address_not_supported | error | invalid_response }` — captured by type
- `heat_pump_calculator_result { annual_bill_delta_median/_p20/_p80, annual_emissions_delta_median, project_type }`
- `heat_pump_pdf_page { page_number }` (every page reached) + `heat_pump_pdf_print`
- already live: `heat_pump_journey_learn_more_click`, `rebate_link_click`, `heat_pump_connect_now_find_installer`, `heat_pump_connect_now_expand_search`

**Views (denominators):**
- `heat_pump_section_view { section: why_heat_pump | bills_impact | find_contractor | rebates | calculator | contractor_pdf }` — fires on render, so click-through rate = clicks / views of the same section.

## Privacy

The calculator collects a street address — its value is **never** sent. `heat_pump_calculator_field { field: 'address' }` records only that the field was engaged. No income/region on these events (Story 4 segmentation is a downstream `screener_uid` join, not FE params).

## Coverage vs. the ticket

- Story 1/3 (clicks): each has a `section_view` denominator for CTR.
- Story 2 (calculator): field / submit / edit / error(by-type) all emitted.
- Story 5/6 (order + correlation): every event carries `screener_uid` (+ GA4 `event_timestamp` / `batch_event_index` for ordering) — reconstructed downstream, no extra FE events.
- Story 7 (savings/emissions trends): `heat_pump_calculator_result` carries the deltas + `project_type` per screening.
- Story 4 (income/region): deferred, downstream join — not FE.

## Tests

114 pass across the heat-pump + PDF-viewer + rebate suites, including new assertions that the card `section_view`s and CTA clicks fire.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added heat-pump journey analytics for section views, CTA clicks, calculator interactions, submissions, results, PDF views, printing, fullscreen, and back navigation.
  * Added privacy-conscious tracking for calculator selections and outcomes without recording street-address values.
  * Added PDF viewer support for page-view, print, and fullscreen tracking callbacks.
* **Tests**
  * Added coverage for heat-pump section views, CTA interactions, calculator result tracking, and PDF page-view callbacks.


src/Assets/analytics/events.ts
```diff
@@ -215,6 +215,81 @@ export interface ScreenerEventMap {
...
+  // human-readable text for the generic API 'error' case.
+  heat_pump_calculator_error: {
+    error_type: 'address_not_supported' | 'error' | 'invalid_response' | 'validation';
+    error_message?: string;
+    field?: string;
+    reason?: string;
+  };
+  // Fires when Calculate impact is pressed, before validation — the true "clicked
+  // the button" count. calculator_submit fires only after validation passes.
+  heat_pump_calculator_submit_attempt: {};
```
david-j-stoll, line 270: Do we need to track whether or not the attempt was successful or failed? Looks like we are tracking an error, but unsure if we need to have the ability to discern whether the error is just a log (error occurred and we're logging that + continuing the process) or a raise to end the workflow. This may be more applicable as we scale the process. Specific workflow I'm thinking of is a failure on the external API. There would be an error, but we wouldn't kill the process.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactResults.tsx
```diff
@@ -191,6 +193,28 @@ export default function CalculateImpactResults({
 
   const equivalencies = getEpaEquivalencies(emMedian);
 
+  const track = useTrackEvent();
```
david-j-stoll, line 196: Looking at useTrackEvent. `context` is set by this line:

`const context = resolveScreenerContext(whiteLabel, uuid, contextWhiteLabel, wrapperContext?.formData?.path);`

`resolveScreenerContext` then sets the screener_state as the whitelabel, with a fallback. However, if the whitelabel was `cesn` there would be no fallback and the screener_state var would be set as cesn. Should this be `co`?

This is pertinent to line 23 at benefits-calculator/src/Assets/analytics/events.ts. I don't think this is breaking anything and it seems like that var should actually be indicating this is a whitelabel value (not a screener state), if anything we may just want to reconsider that variable name. Just something to look at.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -163,16 +164,31 @@ export default function CalculateImpactPage() {
...
   const showCalculateImpact = useFeatureFlag('cesn_heat_pump_journey');
   const [submitState, setSubmitState] = useState<SubmitState>({ status: 'idle' });
+  const track = useTrackEvent();
+  // Guards the address field event so re-selecting/re-typing an address doesn't
+  // refire it — we only want to record that the field was engaged, once.
+  const hasTrackedAddressFieldRef = useRef(false);
+
+  // Reaching the calculator: the denominator for the field and submit events.
+  useEffect(() => {
+    track('heat_pump_section_view', { section: 'calculator' });
```
david-j-stoll: I agree with CodeRabbit on this one, looks like the feature flag conditional for  `showCalculateImpact` is exiting the function after this fires (on line 213). The event would fire even if someone doesn't have the FF toggled on. May want to match what we're doing on the Rebate page: https://github.com/MyFriendBen/benefits-calculator/pull/2145/changes#diff-984671881c80d6488e08835cc1fd7b5ba04957657cd837fad755dad842c6d756R50

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -163,16 +164,31 @@ export default function CalculateImpactPage() {
   const isAdminView = useMemo(() => searchParams.get('admin') === 'true', [searchParams]);
   const showCalculateImpact = useFeatureFlag('cesn_heat_pump_journey');
   const [submitState, setSubmitState] = useState<SubmitState>({ status: 'idle' });
+  const track = useTrackEvent();
+  // Guards the address field event so re-selecting/re-typing an address doesn't
+  // refire it — we only want to record that the field was engaged, once.
```
david-j-stoll: This is interesting. Is there a reason why we don't want to track someone clicking into the field multiple times?

src/Assets/analytics/events.ts
catonph: I think if there's an error the calculator won't run, right? What's an example of an error on the heat pump calculator where then the user still gets the results?

src/Components/Common/PagedDocumentViewer/PagedDocumentViewer.tsx
```diff
@@ -38,8 +53,24 @@ export default function PagedDocumentViewer({ pageImages, pdfUrl, title, classNa
...
+  }, [onPageView]);
+  const goNext = useCallback(() => {
+    setPageIndex((i) => {
+      const next = Math.min(pageCount - 1, i + 1);
+      if (next !== i) {
+        onPageView?.(next + 1);
+      }
+      return next;
+    });
+  }, [pageCount, onPageView]);
```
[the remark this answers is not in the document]
catonph: Good catch — fixed in b3d1599. Moved the onPageView call out of the setPageIndex functional updater in both goPrev/goNext; the updater is now pure and the callback fires once from the event handler after computing the target page. Added a test asserting exactly one page-view per Next/Previous turn.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
```diff
@@ -163,6 +164,16 @@ export default function CalculateImpactPage() {
...
+  const hasTrackedAddressFieldRef = useRef(false);
+
+  // Reaching the calculator: the denominator for the field and submit events.
+  useEffect(() => {
+    track('heat_pump_section_view', { section: 'calculator' });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
catonph: Fixed in b3d1599 — the calculator section_view effect is now gated on showCalculateImpact, so no view is emitted when the feature flag is off (the component returns null in that case). David flagged the same thing.

src/Components/Common/PagedDocumentViewer/PagedDocumentViewer.tsx
```diff
@@ -74,6 +109,13 @@ export default function PagedDocumentViewer({ pageImages, pdfUrl, title, classNa
     [handlePagingKey],
   );
 
+  // Report page 1 once on mount — the "document opened" signal, so callers have a
+  // denominator for later page turns.
+  useEffect(() => {
+    onPageView?.(1);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
catonph: Fixed in b3d1599 — page 1 is now reported only once the page list is non-empty (guarded on pageCount > 0 via a ref, so it fires once when the first non-empty list arrives). Added a test that an empty document emits no page view.

src/Components/EnergyCalculator/Results/RebatePage.tsx
```diff
@@ -142,18 +154,39 @@ function RebateCard({ rebate, rebateCategory }: RebateProps) {
...
+                program: rebate.program,
+                rebate_type: rebate.payment_methods.join(', '),
+                rebate_category: rebateCategory.type,
+                url: rebateUrl,
+              })
+            }
+          >
+            <FormattedMessage id="energyCalculator.rebatePage.applyButton" defaultMessage="Learn how to apply" />
+          </a>
+        ) : (
```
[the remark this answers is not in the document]
catonph: Fixed in b3d1599 — RebateCard now gates heat_pump_rebate_link_click on the same predicate as the journey (useFeatureFlag(cesn_heat_pump_journey) AND hvac). An HVAC rebate shown without the journey flag falls back to the generic outbound_click contract.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
catonph: Agreed — fixed in b3d1599. The calculator section_view effect is now gated on showCalculateImpact so it does not fire when the feature flag is off, matching the pattern on the Rebate page.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactPage.tsx
catonph: Good call out. Doesn't hurt to count every interaction - was thinking we probably wouldn't visualize that (it wasn't on their list) but rather over collect than under collect.

src/Components/EnergyCalculator/Results/HeatPumpJourney/CalculateImpactResults.tsx
catonph: You're right on both counts — it's not breaking anything, and screener_state is really carrying the white-label value (so cesn is correct here, not co).

 Agreed the name is a bit misleading; but resolveScreenerContext / screener_state is shared with the whole screener_* pipeline, so renaming it is probably out of scope for this PR.

src/Assets/analytics/events.ts
david-j-stoll: For this particular workflow we will still get a response, which will inform the error message. But I think this makes sense to me, especially if the goal is to get the submit attempt captured then any errors captured. Looks good.

david-j-stoll (approved):
Approved

